PR #1, opened 2020-10-21
added readme file

nicolasbelfis:


README.md
```diff
@@ -0,0 +1,81 @@
...
+a page using the template you modified  
+The hubspot code is visible in Marketing -> files and templates -> Design tool  
+
+
+### Staging
+
+The next step is to push changes to staging account to allow sales and marketing people double check the impact made by your changes
+
+1. `git push origin <your-branch>`
+2. create a merge request to the `stage branch`
```
ewansheldon, line 62: wouldn't this be less confusing if this was called `staging`

also i think only the `branchname` should be in code block like:

2. create a merge request to the `staging` branch

README.md
```diff
@@ -0,0 +1,81 @@
...
+1. `git push origin <your-branch>`
+2. create a merge request to the `stage branch`  
+
+Once the merge request is accepted, a git action is triggered to upload your changes to hubspot staging account  
+
+**TODO** put link of staging account
+
+### Go live
+
+Once the changes has been accepted create a merge request to `master branch`
```
ewansheldon, line 70: `master` branch?

README.md
```diff
@@ -0,0 +1,81 @@
...
+`npm install`
+
+Connect your local environment to your [developer sandbox](#developer-sandbox) :
+
+1. `npx hs init`
+2. On the prompt press enter, it opens a browser new page
+3. In the page, choose your developer sandbox account and click on **Continue with this account**
+4. copy the personal access key and paste it in your terminal
+5. enter a name that references your sandbox account : this name will be only visible by you
+6. check the file hubspot.config.yml has been created
```
ewansheldon, line 24: maybe put `hubspot.config.yml` in code block (but i'm not sure about that - that's just my preference)

ewansheldon (changes requested):
generally good job man - the instructions were really easy to follow as i set up my sandbox

Morinslash (commented):
I think it would be good to mention also that files removed from the repo are not removed from the HubSpot Accounts and they have to be removed manually inside porta

Adding links to documentation also will be good (CLI commands etc)

README.md
nicolasbelfis: not sure either, if the file is created it is done by the hs process. But maybe we can later one point people to the doc that explains that file ?

ewansheldon (approved):
looks good to me now

Morinslash (approved):

PR #12, opened 2020-11-12
Rename Component

Dan-Bird:
There is an existing component that mimics the Blog Banner caleld `Find out more`. So far this is only used on one page which is the Home page. 

I though it best to rename this component since it's already set to be used within pages and is a copy of the blog banner. I think it might be good to keep these separate, as we can then more easily tailor changes that need to happen within the blog and on the main site.

Once this is merged into staging, we will need to delete the old `Find out more` component. 

The live site is up to date. Nothing needs to be deleted there. I did the rename through Hubspot on the `Codurance` account as the renames are easier there when you're able to see the dependencies. This PR and the one to Master will keep our dev environments up to date.


Morinslash (changes requested):
I think it might be a good idea to refactor also CSS classes. Inside module they are named after old name convention - Find Out More. Shouldn't we also change them to keep current code consistent?

Dan-Bird:
Yeah I think you're

> I think it might be a good idea to refactor also CSS classes. Inside module they are named after old name convention - Find Out More. Shouldn't we also change them to keep current code consistent?

You're absolutely right. This was my bad for missing this. 

I've updated the CSS here: https://github.com/codurance/site-hubspot-migration/pull/12/commits/5cef5bf69b54508239e4f9e4d6701a7bc4f79e94

nicolasbelfis (approved):

Morinslash (approved):

PR #22, opened 2020-11-23
Add Case study collection component

Dan-Bird:
This PR aims to address the [following issue](https://codurance-online.leankit.com/card/1398185072).

In order to keep all the cards the same height, regardless of their content, we need to be able to set a property on the card's container. 

The current cards are individual modules, and their container is just a row within the template which we cannot directly access. 

This module provides a container and an interface in which a user can add cards to it all within the one module.


ewansheldon (approved):
nice work 👍 looks good
feels more straightforward to use too, adding each case study as a property in a list of a larger component

PR #24, opened 2020-11-23
Fix Blog Width

Dan-Bird:
This PR fixed the [following issue](https://codurance-online.leankit.com/card/1393844265).

There were overwriting styles incorrectly setting the blog content's width. I have removed this styles. 

I've wrapped the remaining style within a media query that follows our small breakpoint within our design system, to enable some padding on the sides on mobile devices. 

This provides a better reading experience on these devices but pulling in text from touching the edges of the screens.


ewansheldon (approved):
looks good 👍

nicolasbelfis (approved):

PR #26, opened 2020-11-25
update github actions and readme file

nicolasbelfis:


README.md
```diff
@@ -49,7 +49,7 @@ Hubspot code is within the version_control folder
 
 When you want to test a change made to any hubspot code you have to upload it to your developer sandbox using Hubspot CLI :
 
-`npx hs upload version_control version_control`
+`hs upload version_control/Codurance_September2020 Codurance_September2020`
```
ewansheldon, line 52: maybe soon we should just remove this version control folder altogther, but i guess that's not a job for this pr

ewansheldon (approved):

README.md
nicolasbelfis: agree with that if we never put code outside codurance_september2020

PR #39, opened 2020-12-16
Remove unused fonts

anguspaterson:
There are two fonts referenced in the head that we are not using. These have been removed to reduce load time and page weight


Dan-Bird (approved):
Seems ok 👍

PR #41, opened 2020-12-22
fix css

nicolasbelfis:


anguspaterson:
Looks to be working now -CSS fix applied

anguspaterson (approved):
Fix is working as expected

PR #47, opened 2021-01-07
Correct menu opening bug

Dan-Bird:
This PR fixes the following [ticket](https://codurance-online.leankit.com/card/1415873736).

The issue was that the loops created within the macro were setting the links within sub-menus to have an `aria-controls` attribute that pointed to existing sub-menu id's.

This meant that when clicking the second link within the Services tab, the navbar would not close. Clicking the Our Story link would cause the Services sub-menu to open.


ewansheldon (approved):
Looks good 👍🏻

PR #48, opened 2021-01-13
changes past events to recent events

ewansheldon:


TomSpencerLondon (approved):
This looks good. Changes Past to Recent events.

PR #50, opened 2021-01-14
Services media

ewansheldon:
you can view this module being tested on [this staging page](http://hubspot-developers-87gur3-8691204.hs-sites.com/test-page?hs_preview=qqQqGlZp-37908062188) under the videos and podcasts heading


Dan-Bird (approved):
Looks Alright. We'll need to breakdown the cardSlider.js at some point.

PR #51, opened 2021-01-18
Card modules editable text

ewansheldon:
This makes all text on the card modules (upcoming events, recent events, services video/podcast cards) editable, so that they can be properly translated into spanish


TomSpencerLondon (approved):
These changes look good.

PR #52, opened 2021-01-19
Refactor Homepage with new Design System rules

Dan-Bird:
This PR introduces some new rules regarding design decisions on type, colour and breakpoints. 

It also uses these new rules within the Homepage.


ewansheldon (approved):
Nice work 🌟 looks good

TomSpencerLondon (approved):
We met and discussed this pull request. It was interesting to hear about:
- use of :root to change original styles on the stylesheet
- passing _buttons-utils.css as utils in the HubL macro
- use of variables such as var(--mid-gray) in _colors.css for reuse throughout the stylesheets

PR #55, opened 2021-01-26
Upcoming events translation text

ewansheldon:
this pr:
-  fixes an issue with the image in the header slice of the events page, in chrome and edge
- adds an optional information block to the upcoming events, about our events being in english (for the spanish page)
- fixes a visual issue with the watch all/listen to all cta in the services media in mobile view
- only allows card swipe tracking from within card track

you should be able to see these tested in the staging account event page and service page


TomSpencerLondon (approved):
We have discussed these changes. This looks good! Thanks.

PR #64, opened 2021-02-04
adds styles for subscription preferences page

ewansheldon:


TomSpencerLondon (approved):
This looks good for the background image on the subscription preferences.

PR #85, opened 2021-02-19
adds cookie for exit popup modal to prevent spamming

ewansheldon:


TomSpencerLondon (approved):
This looks good! We have discussed the expiry for the cookie and may alter the cookie expiry depending on UX / business requirements.

PR #89, opened 2021-02-23
Correct Post-migration Defects

Dan-Bird:


version_control/Codurance_September2020/css/modules/try-our-compass-tool.css
```diff
@@ -43,7 +43,7 @@
 
 {% call utils.small() %}
     .cm-compass-teaser [class*="button"] {
-        width: 100%;
+        display: block;
```
TomSpencerLondon, line 46: Nice! This means that the div takes up the whole line.

TomSpencerLondon (approved):

version_control/Codurance_September2020/css/modules/try-our-compass-tool.css
Dan-Bird: Yeah that's correct. For some reason adding the width was making the button extend beyond it's parent's bounds. This way it will just fill the width of it's parent by default.

PR #91, opened 2021-02-23
fill available with header slice image for chrome

ewansheldon:
If you open the [event page](https://www.codurance.com/events/) in chrome, the image is super small
This change makes the image fill the space


TomSpencerLondon (approved):
This is in order to set width for different browsers.

PR #94, opened 2021-02-24
Newsletter enhancements

ewansheldon:
This PR:
- adds a serverless function, which uses the `hubspotutk` cookie to check if the current visitor is a contact
- returns a simple true/false object on if they are a contact
- uses the given serverless function result to decided whether or not to show the exit popup to the current visitor


Dan-Bird (approved):
This looks good.

PR #95, opened 2021-02-25
Benefits module

ewansheldon:
There was a [request to update the company benefits on the become a craftsperson page](https://codurance-online.leankit.com/card/1530818408)

However, this page is currently made up of hubspot drag and drop columns/rows, with each card a "benefits we offer" module

When you try to remove one or two cards, and we no longer have multiple of 3 cards, the cards fill up to take up half/whole width

This change:
- creates a new module which manages the layout of all cards together (Benefits)
- allows you to add a list of benefits to the module via a field
- each benefit takes a heading, icon class name, text
- Removes rich text that currently makes up the text "benefits we offer module"

This create a less fragile and more editable module, and ensures that the layout will be consistent.

We can remove the old "benefits we offer" module after this is merged, and the page is updated with the new module and data

Screenshot 2021-02-25 at 15 03 45


Dan-Bird (approved):
Looks good, LGTM 👍

PR #126, opened 2021-04-20
Homepage banner update ab testing

stefi23:
Added additional styling to support A/B testing on the header. Specifically to have one with a video and another with image only. Once resolved, this section needs to be refactored.


RodrigoCastroS (commented):
well done, we might refactor this upon the A/B testing is finished.

RodrigoCastroS (approved):

PR #128, opened 2021-04-21
Our people header update

RodrigoCastroS:
Adapted a header slice module to the newer headers re-design of Codurance website. Also changed some CSS styles so the cards align with the header title.


stefi23 (approved):
Look great!

PR #129, opened 2021-04-23
Changing the style for small screens

RodrigoCastroS:
The idea was to refactor the layout of the module to use it with the newly design headers


stefi23 (approved):
Looks great!

PR #141, opened 2021-05-25
Added location based date and time  with JS

stefi23:


version_control/Codurance_September2020/modules/Community-events-listing.module/module.html
```diff
@@ -100,10 +100,11 @@ <h2 class="card-slider__heading responsive-page-width">
               <div class="card-slider-item__text-container-inner truncated">
                 <header>
                   <time
-                    class="card-slider-item__metadata"
+                    class="card-slider-item__metadata commmuity-events"
                     datetime="{{ event.datetime|datetimeformat('%Y-%m-%dT%H:%M:%S') }}"
+                    data-time=" {{ event.datetime }}"
```
Dan-Bird, line 105: Do we need two `datetime` attributes? If I'm assuming correctly we could replace the old, `datetime="{{ event.datetime|datetimeformat('%Y-%m-%dT%H:%M:%S') }}"` with the newer `data-time=" {{ event.datetime }}"`

version_control/Codurance_September2020/modules/Community-events-listing.module/module.js
```diff
@@ -14,3 +14,18 @@ function init() {
...
+
+  const timeOptions = { hour12: false, hour: '2-digit', minute:'2-digit', timeZoneName: 'long'};
+  let localTime = new Date(parseInt(timeToConvers)).toLocaleTimeString("en-US", timeOptions)
+  
+  const dateOptions = { weekday: 'short', year: 'numeric', month: 'short', day: 'numeric' };
+  let localDate = new Date(parseInt(timeToConvers)).toLocaleDateString("en-US", dateOptions )
+  localDate = localDate.replace(/,/g, ' ');
+
+  item.innerHTML = `${localDate} | ${localTime} `;
+})
```
Dan-Bird, line 31: Just a quick thought, do you think it makes sense to wrap this in a function, and then call that function within the `init` function defined above - this way the function will be called after the page is loaded and just makes sure that the html is there before we try to select it.

version_control/Codurance_September2020/modules/Upcoming Events Listing.module/module.html
```diff
@@ -99,11 +99,12 @@ <h2 class="card-slider__heading responsive-page-width">
           <div class="card-slider-item__text-container-inner truncated">
             <header>
               <time
-                class="card-slider-item__metadata"
+                class="card-slider-item__metadata upcoming-events"
                 datetime="{{ event.datetime|datetimeformat('%Y-%m-%dT%H:%M:%S') }}"
+                data-time=" {{ event.datetime }}"
```
Dan-Bird, line 104: Same as above

version_control/Codurance_September2020/modules/Upcoming Events Listing.module/module.js
```diff
@@ -14,3 +14,18 @@ function init() {
...
+
+  const timeOptions = { hour12: false, hour: '2-digit', minute:'2-digit', timeZoneName: 'long'};
+  let localTime = new Date(parseInt(timeToConvers)).toLocaleTimeString("en-US", timeOptions)
+  
+  const dateOptions = { weekday: 'short', year: 'numeric', month: 'short', day: 'numeric' };
+  let localDate = new Date(parseInt(timeToConvers)).toLocaleDateString("en-US", dateOptions )
+  localDate = localDate.replace(/,/g, ' ');
+
+  item.innerHTML = `${localDate} | ${localTime} `;
+})
```
Dan-Bird, line 31: Same as above

Dan-Bird (approved):
Looks good. 👍

version_control/Codurance_September2020/modules/Community-events-listing.module/module.js
stefi23: Good idea :)

version_control/Codurance_September2020/modules/Community-events-listing.module/module.html
stefi23: Yes, one is used for the db to get the newest events and the other has to be unix so it can be converted with JS.  I've looked into it too but couldn't find a work around.

PR #142, opened 2021-05-27
Draft page for the newsletter signup with a bit of styling

stefi23:


RodrigoCastroS (approved):
All good! It seems you have an older version of the master branch so it might overwrite the previous fixes they've been done

RodrigoCastroS (approved):
All good! It seems you have an older version of the master branch so it might overwrite the previous fixes they've been done

PR #148, opened 2021-06-04
Lp clients block

stefi23:


version_control/Codurance_September2020/modules/LP-clients-block.module/module.html
```diff
@@ -0,0 +1,33 @@
...
+  <div class="lp-partners-clients-container__content">
+    <h2 class="lp-partners-clients-container__title"> 
+      {% inline_text field="title" value="{{ module.title }}"  no_wrapper=true %}
+    </h2>
+
+    <p class="lp-partners-clients-container__subtitle">
+      {% inline_text field="subtitle" value="{{ module.subtitle }}" no_wrapper=true %}
+    </p>
+  
+    <p class="lp-partners-clients-container__cta">
```
RodrigoCastroS, line 13: I don't know if I would wrap a CTA in a p tag. Maybe a div?

RodrigoCastroS (changes requested):
I left you a comment related to a tag

version_control/Codurance_September2020/modules/LP-clients-block.module/module.html
stefi23: I changed it to a div tag. I was thinking to put an a tag but then it will style it maybe as a link.

RodrigoCastroS (approved):
All good

PR #149, opened 2021-06-04
updated with semantic tags

stefi23:


RodrigoCastroS (approved):

RodrigoCastroS (changes requested):
I leave this as an open PR, because we might merge this to the skeleton branch so we have all the changes there and minimize the error on what we have in each branch.

version_control/Codurance_September2020/templates/landing-page-template.html
```diff
@@ -5,6 +5,12 @@
 -->
 {% extends "./layouts/base.html" %}
 
+{% if locale == 'en' %}
```
RodrigoCastroS, line 8: Shoudln't we do it the other way around?

RodrigoCastroS (approved):

PR #158, opened 2021-06-22
Add logic to select two columns

RodrigoCastroS:
Added the option to choose two columns for the rich text


stefi23 (approved):
Looks great!

PR #163, opened 2021-07-07
Modify module/lp form two columns

RodrigoCastroS:
The design team requested to have a way to choose between two gradient backgrounds for the form module. Giving the opportunity to re-use the module creating Landing pages for the case studies sign up.

### DIFF
--- a/version_control/Codurance_September2020/css/modules/LP-form-two-columns.css
+++ b/version_control/Codurance_September2020/css/modules/LP-form-two-columns.css
@@ -1,12 +1,24 @@
 {% import '../utils/utils.css' as utils %}
 
 .form__wrapper{
-    {{utils.white_to_tango()}}
-    padding: 2em 1.5em;
-    display: flex;
-    flex-direction: column;
-    align-items: center;
+  padding: 2em 1.5em;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  }
+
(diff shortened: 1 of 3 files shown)


version_control/Codurance_September2020/css/modules/LP-form-two-columns.css
```diff
@@ -1,12 +1,24 @@
...
-    flex-direction: column;
-    align-items: center;
+  padding: 2em 1.5em;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  }
+
+.form__wrapper--background-light{
+    {{utils.white_to_tango()}}
```
Dan-Bird, line 11: If I understand correctly, this util is a white to tango gradient.

Just as a thought, it might be a useful consideration to prefix `gradient` or similar. Without any context it's not 100% clear what this is. It could be a transition for example.

Dan-Bird (approved):
Looks good 👍

version_control/Codurance_September2020/css/modules/LP-form-two-columns.css
RodrigoCastroS: Yes, I 100% agree. I might refactor that and maybe passing the colours as a parameter for the macro function. I'll have a look and try to make it work.

PR #164, opened 2021-07-14
Page/case study template

RodrigoCastroS:
Currently, we had problems when using a form in a module as a field, not responding to the redirection of the pages and also not showing the follow-up email checkbox on the Hubspot UI builder. 
Therefore I had to create a template with the same elements to meet the requirements of the team and Hubspot to make it work.


Dan-Bird (approved):
Looks Good 👍

PR #165, opened 2021-08-10
New homepage

RodrigoCastroS:
As part of our brand redesign, we've decided to start to re-build the homepage. This is [the design](https://xd.adobe.com/view/68ba17c6-81d2-4170-8fe1-f0efa69b6c2f-3de9/) 

Most of the modules were created from scratch, or duplicating existing ones and adapting them to the non-drag and drop structure we want to implement in most of the website pages.

### DIFF
--- a/version_control/Codurance_September2020/css/_header.css
+++ b/version_control/Codurance_September2020/css/_header.css
@@ -9,7 +9,7 @@ You will also find the mobile menu styles below.
 {% include "../utils/utils.css" as utils %}
 
 header.header {
-  min-height: 90px;
+  min-height: 85px;
 }
 .cm-header-group {
   background-color: white;
--- a/version_control/Codurance_September2020/css/_header.css
+++ b/version_control/Codurance_September2020/css/_header.css
@@ -6,6 +6,8 @@ You will also find the mobile menu styles below.
 *********************************************************************************
 ****************************************************************************** */
 
+{% include "../utils/utils.css" as utils %}
+
 header.header {
   min-height: 90px;
 }
@@ -17,6 +19,7 @@ header.header {
   transition: transform 0.3s ease;
   width: 100%;
   z-index: 100;
+  {{ utils.card_shadow() }}
 }
 .custom-logo img {
   max-width: 147px;
@@ -74,6 +77,7 @@ header.header.fixed  .cm-header-group {
 #nav-toggle {
   display: none;
 }
+
 
 /* Styles for the Logo */
 
@@ -309,16 +313,7 @@ header.header .mobile-logo{
     padding-left: 50px;
     padding-right: 50px;
   }
-  .cm-header-group:after{
-    background-color: #b7b7b7;
-    content: "";
-    height: 1px;
-    margin-top:1px;
-    margin-left: auto;
-    margin-right: auto;
-    max-width: var(--max-page-width);
-    display: block;
-  }
+ 
 }
 
 @media (min-width: 1300px){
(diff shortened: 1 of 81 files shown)


version_control/Codurance_September2020/css/_header.css
```diff
@@ -309,16 +313,7 @@ header.header .mobile-logo{
...
-    margin-top:1px;
-    margin-left: auto;
-    margin-right: auto;
-    max-width: var(--max-page-width);
-    display: block;
-  }
+ 
 }
 
 @media (min-width: 1300px){
```
Dan-Bird, line 319: Do any of the media queries in this file align with the utility ones we have? 

If they do this could be a good opportunity to get them unified and using the shared media queries.

version_control/Codurance_September2020/css/modules/compass.css
```diff
@@ -0,0 +1,88 @@
...
+    {{ utils.freed() }}
+   font-weight: var(--light-font-weight);
+}
+
+.compass-content__text p{
+    {{ utils.natus() }}
+     margin-top: 1em;
+}
+
+.compass-content__expetise-areas {
```
Dan-Bird, line 29:
```suggestion
.compass-content__expertise-areas {
```

version_control/Codurance_September2020/css/modules/recent-posts-v2.css
```diff
@@ -0,0 +1,162 @@
+{% import '../utils/utils.css' as utils %}
+
+body{
```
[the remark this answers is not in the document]
RodrigoCastroS: 100% true. By mistake I misplaced this rule which it has to be in the homepage-v2.css So in that case I believe it will only affect to that page. Angus wanted to move into the full width bg approach and was the solution I found without compromising the whole website.

version_control/Codurance_September2020/css/modules/recent-posts-v2.css
```diff
@@ -0,0 +1,162 @@
...
+  max-width: none;
+}  
+
+.cm-recent-post {
+  max-width: 1200px;
```
Dan-Bird, line 8: This max width property also appears within the compass module's styles. 

If there are any other modules that are using this, I wonder if it will be beneficial to create a global width variable that could be used. This way any edits or changes can be kept in synch across the modules that rely on the width.

version_control/Codurance_September2020/css/modules/software-craftsmanship-v2.css
```diff
@@ -0,0 +1,165 @@
...
+  top: 0;
+}
+
+{% call utils.medium_large_and_extra_large() %}
+
+.software-craft {
+  padding: 2.5em;
+}
+
+{% endcall %}
```
Dan-Bird, line 164: Maybe this block would be better located at the top of the file under the `.software-craft` block.

Dan-Bird (approved):
Looks good 👍

version_control/Codurance_September2020/modules/Compass.module/module.html
```diff
@@ -0,0 +1,35 @@
+
+{{ require_css(get_asset_url("../../css/modules/compass.css")) }}
+
+<div class="compass-wrapper">
+	<div class="compass-content">
+		<div class="compass-content__text">
+			{% inline_rich_text field="richtext_field" value="{{ module.richtext_field }}" %}
+		</div>
+		<div class="compass-content__expetise-areas">
```
Dan-Bird, line 9:
```suggestion
		<div class="compass-content__expertise-areas">
```

version_control/Codurance_September2020/css/modules/compass.css
RodrigoCastroS: What an eye to spot this one! Thanks

version_control/Codurance_September2020/css/modules/software-craftsmanship-v2.css
RodrigoCastroS: Done! I moved it to the right place

version_control/Codurance_September2020/css/modules/recent-posts-v2.css
RodrigoCastroS: I've set up a max-with on the root, as it was there declared the max-page-width before

PR #171, opened 2021-09-28
Change the overlay position to cover side to side

RodrigoCastroS:
There was a bug related to the overlay on the modal in our people page. More info on the issue here (https://codurance-unit.monday.com/boards/1559115273/pulses/1709798346)


Dan-Bird (approved):
All good 👍

PR #172, opened 2021-09-28
Feature/landing-page-modules

RodrigoCastroS:
The marketing team has requested some changes into the Landing pages modules:
* Have an optional regular button link so they can use an anchor link on it. 
* The form to be centered when there isn't any content on the left


Dan-Bird (approved):
All looks good 👍

PR #174, opened 2021-11-09
Update/new careers page

RodrigoCastroS:
Update the current careers page realigning the modules and change the order of appearance. There is also a new Locations module on the page that was developed on the new/job-listing page branch.

### DIFF
--- a/version_control/Codurance_September2020/css/pages/careers-v2.css
+++ b/version_control/Codurance_September2020/css/pages/careers-v2.css
@@ -100,6 +100,7 @@ body{
     max-width: 50em;
 }
 
+
 /*------------------------------------*\
   #Benefits Section
 \*------------------------------------*/
@@ -217,11 +218,28 @@ body{
     color: var(--tango-400);
 }
 
+.craftspeople .text-and-image__copy ol{
+    list-style: revert;
+}
+
+.craftspeople ol li{
+    margin-bottom: 0.5em;
+    margin-left: 1em;
+}
+
 .craftspeople .cta-button.button-secondary{
     background: white;
     color: var(--persimmon);
 }
 
+.craftspeople .text-and-image__image {
+    scale: 1.2;
+    align-self: start;
+}
+
+.craftspeople .text-and-image__image img {
+ transform:translate(-20%);
+}
 
 
 {% call utils.small() %}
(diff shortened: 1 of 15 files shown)


version_control/Codurance_September2020/css/pages/careers-v2.css
```diff
@@ -86,6 +86,21 @@ body{
     }
 {% endcall %}
 
+/*------------------------------------*\
+  #Locations Section
+\*------------------------------------*/
+
+.locations{
+    background: linear-gradient(180deg,hsla(24.2,86.3%,54.3%,0.3) 0%,white 100%);
```
Dan-Bird, line 94: Are these one off colour values? If we require this value in other areas then we could add it to the design system.

version_control/Codurance_September2020/css/pages/careers-v2.css
```diff
@@ -86,6 +86,21 @@ body{
...
+}
+
+.locations_container{
```
Dan-Bird, line 97: Should this be a double underscore or a hyphen?

version_control/Codurance_September2020/templates/careers-v2.html
```diff
@@ -50,10 +50,25 @@ <h1>Careers</h1>
...
   </section>
+  <section class="craftspeople">
+    {% module 'craftspeople' path='../modules/Text-and-image.module' label="Craftspeople section" no_wrapper=True%}
+  </section>
+  <section class="testimonials">
+    {% module 'Testimonial Slider' path='../modules/Landing Page Modules/LP-testimonial-slider.module' label="Testimonial Slider"  no_wrapper=True %}
+  </section>
 
+  <section class="locations">
+    <div class="locations_container">
```
Dan-Bird, line 61: Similar to the CSS, should this be a double underscore or a hyphen?

Dan-Bird (approved):
👍

version_control/Codurance_September2020/css/pages/careers-v2.css
RodrigoCastroS: double underscore. Thanks!

version_control/Codurance_September2020/templates/careers-v2.html
RodrigoCastroS: double underscore. Thanks!

version_control/Codurance_September2020/css/pages/careers-v2.css
RodrigoCastroS: It is... kind of. When designers are creating gradients, sometimes they use different directions and in this case, they might use the base colour (in this case persimmon) but with different hue and alpha values. So I don't know if I would be able to change the value of a colour property value coming from a variable.

PR #175, opened 2021-11-16
Fix/uptated blog post date

RodrigoCastroS:
We found an issue with the updated dates. It seems that most of our blog posts were added on a certain date so a small change would signify that the post was updated. The solution was to create a reference date (September 1st, 2021) to be compared with the posts that were updated.


Dan-Bird (approved):
LGTM 👍

PR #183, opened 2022-02-01
Update/testimonials-module

RodrigoCastroS:
Given the current situation of the module, I've created two new editable fields (a title and a button). The slider works even if there are empty fields. The way is built might cause issues when adding a background (as it was created only to be used on landing pages as a whole section), so probably it might need a more simplified version and adding a simple image background from HubSpot sections.


Dan-Bird (approved):
Looks good 👍

PR #186, opened 2022-02-21
Update-case-study-card-collection

RodrigoCastroS:
Update design current case study cards for service line pages beign part of the redesign og the service line pages


Dan-Bird (approved):
Seems sensible 👍

PR #188, opened 2022-02-22
Adjustments to LP Testimonial Slider Module

vicarali:
This module has been configured in order to apply the following changes to the slider when it just contains a single slide:

- Don't display the horizontal navigation arrows
- Disable the slider (specifically, the swiper slider object itself will not be created)

### DIFF
--- a/version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.html
+++ b/version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.html
@@ -46,10 +46,6 @@
           {% if module.testimonial_items|length > 1 %}
             <div class="swiper-button-next"></div>
             <div class="swiper-button-prev"></div>
-          {% else %} 
-            <script>
-              var swiperOneSlide = true;
-            </script>
           {% endif %}
 
           <div class="swiper-pagination"></div>
--- a/version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.html
+++ b/version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.html
@@ -42,8 +42,16 @@
             </div>
           {% endfor %}
           </div>
-          <div class="swiper-button-next"></div>
-          <div class="swiper-button-prev"></div>
+
+          {% if module.testimonial_items|length > 1 %}
+            <div class="swiper-button-next"></div>
+            <div class="swiper-button-prev"></div>
+          {% else %} 
+            <script>
+              var swiperOneSlide = true;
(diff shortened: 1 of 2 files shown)


version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.html
```diff
@@ -42,8 +42,16 @@ <h2 class="testimonial-slider__title">{{ module.header.title }}</h2>
...
           {% endfor %}
           </div>
-          <div class="swiper-button-next"></div>
-          <div class="swiper-button-prev"></div>
+
+          {% if module.testimonial_items|length > 1 %}
+            <div class="swiper-button-next"></div>
+            <div class="swiper-button-prev"></div>
+          {% else %} 
+            <script>
```
RodrigoCastroS, line 50: Isn't there a better way without injecting a script on the body? Maybe accessing it through the DOM could be a way to avoid this.

version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.html
```diff
@@ -42,8 +42,16 @@ <h2 class="testimonial-slider__title">{{ module.header.title }}</h2>
...
+              var swiperOneSlide = true;
```
RodrigoCastroS, line 51: In case we keep this solution, I might use let, because you can't redeclare the variable without being explicit. Another way is using a constant, it seems the value of the boolean it won't change

version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.js
```diff
@@ -1,9 +1,11 @@
-
-
-var swiper = new Swiper(".swiper", {
+/* The swiper slider of the module will only be created
+if there is more than one slide */
+if(typeof swiperOneSlide == "undefined"){
```
RodrigoCastroS, line 3: You could use triple equals ("===") to compare, it's a more strict aproach and it might avoid errors

RodrigoCastroS (changes requested):
Overall all good! But check the comments I left you

version_control/Codurance_September2020/modules/Landing Page Modules/LP-testimonial-slider.module/module.js
vicarali: True! Corrected for the next commit

vicarali:
@RodrigoCastroS I've added a new commit with the new features refactored in order to avoid the usage of JS code on the HTML file 👍

PR #190, opened 2022-03-16
New template/service line hub

vicarali:
Creation of template for the renovated Service Line Hub Page with the following changes:

- Addition of a Hub Menu section to navigate between the different Codurance Services
- Adjustments to the Clients section/module
- Addition of an Industries section to display the different industries Codurance has worked on
- Creation of a section/module about Compass to substitute the older Compass widget
- Adjustments to the Contact Form section

### DIFF
--- a/version_control/Codurance_September2020/css/modules/links-hub.css
+++ b/version_control/Codurance_September2020/css/modules/links-hub.css
@@ -59,3 +59,19 @@
     max-width: var(--eames-max-width);
 }
 
+
+{% call utils.medium_large_and_extra_large() %}
+
+    
+    .hub-menu{
+        position: relative;
+        height: 25em;
+        display: grid;
+        place-items: center;
+    }
+    .hub-menu__item{
+        position: absolute;
+    }
+    .hub-menu__item{
+    }
+{% endcall%}
--- a/version_control/Codurance_September2020/css/modules/links-hub.css
+++ b/version_control/Codurance_September2020/css/modules/links-hub.css
@@ -5,7 +5,7 @@
     {% set angle = 360 / item_count|length %}
 
     {% for item in item_count %}
-        {% if loop.index == 2 or loop.index == 5 %}
+        {% if loop.index == 2 %}
             {% do rotation.update({degrees_correction : rotation.degrees + 45}) %}
         {% else %}
             {% do rotation.update({degrees_correction : rotation.degrees}) %}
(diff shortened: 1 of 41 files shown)


version_control/Codurance_September2020/modules/Icon-cloud.module/module.html
```diff
@@ -0,0 +1,21 @@
...
+			{% if icon.image.size_type == 'auto' %}
+				{% set sizeAttrs = 'style="max-width: 100%; height: auto;"' %}
+			{% elif icon.image.size_type == 'auto_custom_max' %}
+				{% set sizeAttrs = 'width="{{ icon.image.max_width }}" height="{{ icon.image.max_height }}" style="max-width: 100%; height: auto;"' %}
+			{% endif %}
+			{% set loadingAttr = icon.image.loading != 'disabled' ? 'loading="{{ icon.image.loading }}"' : '' %}
+			<div class="icon">
+				<img class="icon__image"
+					 src={{ icon.image.src }} 
+					 alt="{{ icon.image.alt }}" {{ loadingAttr }}>
```
RodrigoCastroS, line 16: I don't know if the quotes between the {{}} would work, also we're not using the size attributes so we might be able to get rid of those

version_control/Codurance_September2020/templates/partials/sections/feature-section.html
```diff
@@ -0,0 +1,24 @@
+<!--
+    templateType: page
+    isAvailableForNewContent: false
```
RodrigoCastroS, line 3: I'm still wondering if we would benefit doing partials sections instead of creating the elements on the page we are going to use. We might be adding complexity folder hierarchy for no reason but we should take a decision to go this route or not and document it

RodrigoCastroS (approved):
Good work! I made some comments that I've already fixed with some commits

version_control/Codurance_September2020/css/modules/links-hub.css
```diff
@@ -0,0 +1,340 @@
...
+.hub-menu__link-wrapper {
+    display: flex;
+    gap: 1em;
+}
+
+.hub-menu__link:hover{
+    text-decoration: none;
+}
+
+.hub-menu__link-wrapper {
```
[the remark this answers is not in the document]
vicarali: I've seen that you already corrected that, thank you!

version_control/Codurance_September2020/modules/Icon-cloud.module/module.html
vicarali: I've just sent another commit to fix this issue of the module by setting quotes on the src attribute for better understanding, cleaning the unused size parameters on the HTML and disallowing their edition on the content editor.

PR #191, opened 2022-03-18
Fix modules routing and module button name

RodrigoCastroS:
We found some routes were creating issues when importing the modules into the partials sections.


vicarali (approved):
Tested again just in case and everything seems perfect now!

PR #193, opened 2022-03-18
Fix/icons cloud path

vicarali:
Fix incorrect paths for modules and CSS assets of the feature section partial


RodrigoCastroS (approved):
Seems good to me 👍

PR #194, opened 2022-03-21
Fix hub menu links and scale on hover animation

vicarali:
- Href attributes weren't loading correctly for these links

- Add transform-scale as fallback to ensure browser compatibility


RodrigoCastroS (approved):
Seems good! thanks for keeping an eye on details ⚡

PR #196, opened 2022-03-25
Fix menu logo redirection url on click to be the right lang

RodrigoCastroS:
The redirection to the homepage when clicking on the logo of the landing pages were always pointing to the '/es' site. We've changed the conditional to fix the issue.


vicarali (approved):
LGTM!

PR #197, opened 2022-03-29
Creation of new landing page template with menu

vicarali:


version_control/Codurance_September2020/templates/landing-page-navigation-template.html
```diff
@@ -0,0 +1,83 @@
...
+  {{ require_js(get_asset_url("../js/main.js")) }}
+  {{ standard_header_includes }}
+  {{ require_css(get_asset_url("../css/FontAwesome.css")) }}
+  {{ require_css(get_asset_url("../css/theme-overrides.css")) }}
+</head>
+
+{% block body %}
+ 
+        <section>
+            {% dnd_area "dbd-main-section" class='section-area-full', label='Main Section' %}
```
RodrigoCastroS, line 36: I think there's a misspelling mistake on the dnd area. Maybe you meant "dnd-main-section" :)

version_control/Codurance_September2020/templates/landing-page-navigation-template.html
```diff
@@ -0,0 +1,83 @@
+<!--
+templateType: page
+isAvailableForNewContent: true
+label: Landing Page Navigation Template
```
RodrigoCastroS, line 4: I would probably change the name of the template so any user can identify better the one they have to use

RodrigoCastroS (approved):
I've added a couple of comments that you might want to have a look. Besides that, it looks good to me 👍

PR #200, opened 2022-03-29
Create/case study template

RodrigoCastroS:
Created a new template for the ungated case studies, where we can continue improving and adding sections to the template.


vicarali (approved):
Tested on staging environment and everything seems fine!

PR #201, opened 2022-03-31
Create/past-events-page-template

RodrigoCastroS:
Create page to showcase all past events


vicarali (approved):
Good job! I've added some additional commits to adjust the flexbox and to refactor the recent events module and the listing section from the past events template.

PR #202, opened 2022-04-05
Feature/events page gated content

RodrigoCastroS:
In order to differentiate from gated to ungated content we changed a bit the style of the cards also the section has changed accordingly

### DIFF
--- a/version_control/Codurance_September2020/modules/Recent-Events-Listing.module/module.html
+++ b/version_control/Codurance_September2020/modules/Recent-Events-Listing.module/module.html
@@ -1,6 +1,7 @@
 {% import '../../snippets/button-snippets.html' as snippets %}
 
 {{ require_css(get_asset_url("../../css/card-slider.css")) }}
+
 {% if module.hubdb_table %} 
   {% set previousEventsFilter = "&orderBy=-date&limit=4" %}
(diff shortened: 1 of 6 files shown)


version_control/Codurance_September2020/modules/Recent-Events-Listing.module/module.html
```diff
@@ -115,12 +118,21 @@ <h4 class="card-slider-item__title">
             </p>
           </div>
           <div class="card-slider-item__button-container">
+          {% if event.video is string_containing 'livestorm' %}
```
vicarali, line 121: I don't know if it would be better to set these different "classnames" and "text" values as variables (along with the condition). This way the "button_primary" snippet call would only be required to be included once. What do you think?

vicarali (changes requested):
I just left a single comment on a part that maybe could be improved (I'm not entirely sure, but please review it before the merge). Other than that, I already tested it on staging and everything looks perfect!

version_control/Codurance_September2020/modules/Recent-Events-Listing.module/module.html
RodrigoCastroS: I've created a small object to contain the info, so it's a bit cleaner when passing the parameters to the button mactro

vicarali (approved):
Re-checked on staging and everything looks perfect! Thank you!

PR #203, opened 2022-04-06
Create/header past events page

vicarali:
- Creation of header section with background
- Refactoring of past events listing section from usage of flexbox to grid
- General style adjustments of the page template


version_control/Codurance_September2020/modules/Recent-Events-Listing.module/module.html
```diff
@@ -136,6 +144,14 @@ <h4 class="card-slider-item__title">
                 target="_blank",
                 text= button.text
             )}}
+          {% else %}
```
RodrigoCastroS, line 147: The same with this button. The conditional was done on line 134, so I don't think this will be needed, since it's rendering one button.

RodrigoCastroS (commented):
Looks good to me. I left you some comments where you might need to change some code. 🔢

RodrigoCastroS (approved):
All good 👍

PR #205, opened 2022-04-19
Feature/youtube-events-popup

RodrigoCastroS:
Create a modal to watch videos instantly on the current and past events page


version_control/Codurance_September2020/modules/YouTube-Video-Popup.module/module.js
```diff
@@ -0,0 +1,65 @@
...
+closeModalButton.addEventListener("click", () => hideModal(overlay) );
+
+document.addEventListener('click', function(event){
+    if(event.target.className !== "overlay") return;
+
+    hideModal(overlay);
+})
+
+document.addEventListener('keyup', function(event){
+    if(event.code !== "Escape") return;
```
vicarali, line 20: I love this functionality, very well thought! 

Even so, personally I think that just using an equal condition on this if statement and executing the function in positive case it would be better practice, to avoid the usage of the return statement (in that case, this line would just be `if(event.code == "Escape") hideModal(overlay);`).

What do you think?

version_control/Codurance_September2020/modules/YouTube-Video-Popup.module/module.js
```diff
@@ -0,0 +1,65 @@
...
+    const regExp = /^.*(youtu\.be\/|v\/|u\/\w\/|embed\/|watch\?v=|\&v=)([^#\&\?]*).*/;    
+    const match = str.match(regExp);
+    const idLength = 11;
+
+    if (match && match[2].length == idLength) return match[2];
+    
+    throw new Error('Expression not found');  
+}
+
+function showModal(elem) {
```
vicarali, line 50: Maybe we could change the name of the "elem" parameter from this function and the following one to something more descriptive, like "overlay". What do you think?

vicarali (approved):
Really good job and implementation! Congratulations :)

I've just added a few comments regarding some possible minor improvements; if you want to discuss anything about them please just let me know!

version_control/Codurance_September2020/modules/YouTube-Video-Popup.module/module.js
RodrigoCastroS: I see what you mean, but how would you call the parameter you pass into the function? The reason I made this was, so this function could be reusable in a way, although I probably have to extract into other functions. In my point of view, having the paramenter and arguments having the same name could create a bit of confusion but I don't mind changing it.

vicarali: Mmmm fair enough, in fact what you are making me think about right now is that maybe these functions don't even require parameters haha, maybe we could just use the constant of the overlay directly. What do you think abou that?

If you don't like that approach either don't worry, we can leave it as is.

PR #208, opened 2022-04-25
Fix/event pages adjustments

vicarali:
- Deletion of overlay feature for gated event cards
- Creation of new URL field for "watch all" CTA at the bottom of the general events page


RodrigoCastroS (approved):
Seems good to me!

PR #216, opened 2022-05-18
Fix/accordion module display issues

vicarali:
- Refactoring of JS code of the module for clarity and performance
- Keep all accordion items closed by default 
- Fix clipping between accordion titles and the toggling icon
- Fix minor issue on the Careers Overview Page Template (specifically the "Why Join Us Section") on which the image was positioned incorrectly when it was displayed at the right

### DIFF
--- a/version_control/Codurance_September2020/css/modules/accordion.css
+++ b/version_control/Codurance_September2020/css/modules/accordion.css
@@ -8,11 +8,11 @@
     padding-right: 50px;
 }
 
-.accordion .panel {
+.panel {
     border-bottom: 1px solid var(--french-gray);
 }
 
-.accordion .panel:last-child {
+.panel:last-child {
     border-bottom: 0;
 }
 
@@ -53,11 +53,11 @@
     transform: rotate(-90deg);
 }
 
-.panel:not(.accordion-close) .panel-title:after {
+.panel:not(.panel--close) .panel-title:after {
     transform: rotate(0);
 }
 
-.panel:not(.accordion-close) .panel-title:before {
+.panel:not(.panel--close) .panel-title:before {
     transform: rotate(0);
 }
 
@@ -66,13 +66,15 @@
     font-weight: var(--heavy-font-weight);
 }
 
-.accordion .panel-collapse {
-    padding-bottom: 40px;
-    display:none;
-}
-
 .panel-body {
     {{ utils.base() }}
+    transition-property: max-height;
+    overflow: hidden;
+    max-height: 0;
+}
+
+.panel-body > .frame {
+    padding-bottom: 2em;
 }
 
 .panel-body p, ul, li {
--- a/version_control/Codurance_September2020/css/modules/accordion.css
+++ b/version_control/Codurance_September2020/css/modules/accordion.css
@@ -1,6 +1,22 @@
 {% import '../utils/utils.css' as utils %}
 
-.panel-heading {
+.accordion {
+    margin-left: auto;
+    margin-right: auto;
+    max-width: 1000px;
+    padding-left: 50px;
+    padding-right: 50px;
+}
+
+.accordion .panel {
+    border-bottom: 1px solid var(--french-gray);
+}
+
+.accordion .panel:last-child {
+    border-bottom: 0;
+}
+
+.panel-title {
     padding-bottom: 20px;
     padding-top: 20px;
     position: relative;
@@ -12,22 +28,13 @@
     display:block;
 }
 
-.panel-heading p {
-    margin: 0;
-}
-
-.accordion-toggle {
-    {{ utils.freed() }}
-    font-weight: var(--heavy-font-weight);
-}
-
-.panel-heading:focus,
-.panel-heading:hover{
+.panel-title:focus,
+.panel-title:hover{
     color: var(--shark);
 }
 
-.panel-heading::before,
-.panel-heading::after {
+.panel-title::before,
+.panel-title::after {
     content: "";
     background-color: var(--persimmon);
     content: '';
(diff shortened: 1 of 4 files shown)


version_control/Codurance_September2020/css/modules/accordion.css
```diff
@@ -1,65 +1,83 @@
...
-    cursor: pointer;
-    margin-top: 0;
-    text-decoration: none !important;
-    margin-bottom: 1rem;
-    color: var(--shark);
-    display:block;
+.accordion {
+    margin-left: auto;
+    margin-right: auto;
+    max-width: 1000px;
```
RodrigoCastroS, line 6: I believe we have a global varibale to replace instead putting manually the withs

version_control/Codurance_September2020/css/modules/accordion.css
```diff
@@ -1,65 +1,83 @@
...
-    {{ utils.freed() }}
-    font-weight: var(--heavy-font-weight);
+.panel:last-child {
+    border-bottom: 0;
 }
 
-.panel-heading:focus,
-.panel-heading:hover{
+.panel-title {
+    --panel-title-icon-width: 1.25em;
```
RodrigoCastroS, line 20: A suggestion here would be, if the css property it's already inside the panel title, maybe just using icon-width it would be more accurate

version_control/Codurance_September2020/css/modules/accordion.css
```diff
@@ -1,65 +1,83 @@
...
+
+    color: var(--shark);
+    cursor: pointer;
+    display: block;
+    margin-bottom: 1rem;
+    margin-top: 0;
+    padding-bottom: 1em;
+    padding-top: 1em;
+    position: relative;
+    text-decoration: none !important;
```
RodrigoCastroS, line 30: I recommend avoiding the use of !important rules. Try to use the cascade. Maybe you could extract this rule into the property hover state and it still work. (.panel-title:hover)

version_control/Codurance_September2020/css/modules/accordion.css
```diff
@@ -1,65 +1,83 @@
...
 
-.panel-heading::after {
+.panel-title::after {
     content: "";
     transform: rotate(-90deg);
 }
 
-.accordion .panel {
-    border-bottom: 1px solid var(--french-gray);
+.panel:not(.panel--close) .panel-title:after {
```
RodrigoCastroS, line 58: since the before and after have the same rules, they could be together on the same property sharing the rule

version_control/Codurance_September2020/modules/Accordion.module/module.js
```diff
@@ -1,88 +1,79 @@
-if (window.NodeList && !NodeList.prototype.forEach) {
-  NodeList.prototype.forEach = Array.prototype.forEach;
+let accordionItems = document.querySelectorAll(".panel");
```
RodrigoCastroS, line 1: I wonder if it's better to use a constant, since we're not modifiying the value of the variable, just looping through the html nodes

version_control/Codurance_September2020/modules/Accordion.module/module.js
```diff
@@ -1,88 +1,79 @@
...
-  target.style.removeProperty('margin-bottom');
-  window.setTimeout( () => {
-    target.style.removeProperty('height');
-    target.style.removeProperty('overflow');
-    target.style.removeProperty('transition-duration');
-    target.style.removeProperty('transition-property');
-  }, duration);
+
+function slideUpSiblingAccordionItems(accordionItem) {
+  slideUpPreviousSiblingAccordionItems(accordionItem);
```
RodrigoCastroS, line 18: Having this much abstraction it's hard to understand what are the previous siblings or the next ones. So it ends up having a really long method names giving the user hard times to read

version_control/Codurance_September2020/modules/Accordion.module/module.js
```diff
@@ -1,88 +1,79 @@
...
+  slideUpNextSiblingAccordionItems(accordionItem);
 }
-var slideToggle = (target, duration = 500) => {
-  if (window.getComputedStyle(target).display === 'none') {
-    return slideDown(target, duration);
-  } else {
-    return slideUp(target, duration);
+
+function slideUpPreviousSiblingAccordionItems(accordionItem) {
+  let previousSiblingAccordionItem = accordionItem.previousElementSibling;
```
RodrigoCastroS, line 23: Still, really long variable names for being an scoped variable IMHO. It seems it's the same length as invoquing the DOM method previousELSibling

version_control/Codurance_September2020/modules/Accordion.module/module.js
```diff
@@ -1,88 +1,79 @@
...
+  accordionItemBody.style.removeProperty('max-height');
+}
+
+function slideDown(accordionItemBody, transitionDuration) {
+  accordionItemBody.style.transitionDuration = transitionDuration + 'ms';
+  const accordionItemBodyHeight = accordionItemBody.scrollHeight;
+  accordionItemBody.style.maxHeight = accordionItemBodyHeight + 'px';
+}
+
+$('.panel-title').click(function(e){
```
RodrigoCastroS, line 76: It seems this is jQuery mixed with javascript. I'm not going to ask you to refactor this, but after all you've gone through this would have been the extra step to get rid of a file. But because I know that we might end up using the native elements of HTML for accordions we might not need even javascript in the end.

RodrigoCastroS (approved):
Great job overall! I've left you some comments to have a look. I can see the effort you've put through refactoring an existing module we had.

version_control/Codurance_September2020/css/modules/accordion.css
vicarali: Spot on! I've changed it to the equivalent custom property

version_control/Codurance_September2020/css/modules/accordion.css
vicarali: Good idea! I just changed it

version_control/Codurance_September2020/css/modules/accordion.css
vicarali: I'm sorry, this slipped under my radar. You were right, just by moving the property to the following block it already gives it sufficient priority, even without the !important :)

version_control/Codurance_September2020/css/modules/accordion.css
vicarali: That's right, I wonder why it was set up that way in the beginning

version_control/Codurance_September2020/modules/Accordion.module/module.js
vicarali: For now we've decided to remove the scrolling functionality on accordion item toggling, so there's no jQuery on this file anymore.

version_control/Codurance_September2020/modules/Accordion.module/module.js
vicarali: Resolved by merging these two old functions.

PR #217, opened 2022-05-20
Create/soft-mod-link-to-pillar

RodrigoCastroS:
Created an optional section on the Service Line header module, to redirect to any page we want but in this case it would be to redirect the user to the pillar pages with its topic.


vicarali (approved):
Great work!! I've tested the module on staging just in case and everything looks perfect!

I've seen that it hasn't been verified if it's compatible with the master branch (pull), it will be required to do so before merging and then it's completely ready to go.

PR #224, opened 2022-06-14
Update/max width website

vicarali:
- Adjustments on several templates of the site to be able to remove the existing max-width limitation of 1800px
- Refactoring and cleaning of the layout and JS code related to the Isotope library used on clients page


RodrigoCastroS (approved):
Great work! 👍  I wanted to do this refactoring on the web for a long time. Let's keep the good work 🎸

PR #227, opened 2022-06-28
Fix/minor adjustments becoming a craftsperson page

vicarali:
- Adjustments to font-size and sizing/positioning of the title of the Video-Mosaic module
- Adjust padding for interview intro section

### DIFF
--- a/version_control/Codurance_September2020/css/modules/video-mosaic.css
+++ b/version_control/Codurance_September2020/css/modules/video-mosaic.css
@@ -92,10 +92,10 @@
 
     display: flex; 
     align-items: center;
+    justify-content: center;
     grid-column: 4/7;
     grid-row: span 2;
     height: calc(100% - 1.5em);
-    padding-right: 2.5em;
 
     /* Center box vertically in relation to first image of mosaic */
     position: relative;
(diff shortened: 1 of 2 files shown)


version_control/Codurance_September2020/css/modules/video-mosaic.css
```diff
@@ -124,9 +130,6 @@
   }
 
   .mosaic-collection__title > h2 {
-    {{ utils.freed() }}
-
-    font-weight: initial;
     width: var(--habib-max-width);
```
RodrigoCastroS, line 133: I wonder if you could use change this width for a max-width instead. Also, since is the width and it could adapt to the container, I guess moving it to line 36 it might work as well

RodrigoCastroS (approved):
LGTM 🚀 I left you a comment on the video mosaic styles to simplify the property declaration

version_control/Codurance_September2020/css/modules/video-mosaic.css
vicarali: You are right, I recall that there was a reason for that before but that doesn't seem to be the case anymore 🤔 

I've just sent a commit to ammend this style and the one for the large and extra_large viewpoints 👍 

Regarding the suggestion to move this style to line 36: unfortunately I think it would not work in this case, because the max-width for the title varies between habib for the medium viewpoint and gaudi for large and extra large.

RodrigoCastroS: gotcha! makes sense though :)

PR #233, opened 2022-07-05
Show upcoming events only if there are any

vicarali:
If there aren't, a padding will be set at the top of the main container for a better separation with the header

### DIFF
--- a/version_control/Codurance_September2020/modules/Upcoming Events Listing.module/module.html
+++ b/version_control/Codurance_September2020/modules/Upcoming Events Listing.module/module.html
@@ -3,138 +3,146 @@
 {{ require_css(get_asset_url("../../css/card-slider.css")) }}
 
 {% if module.hubdb_table %}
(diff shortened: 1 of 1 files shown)


version_control/Codurance_September2020/modules/Upcoming Events Listing.module/module.html
```diff
@@ -3,138 +3,146 @@
...
-        {% form "email_form" form_to_use={{ module.newsletter_form.form_id }} %}
+
       </div>
     </div>
-
-  </div>
-</div>
-
+  {% else %}
+    {% require_css %}
```
RodrigoCastroS, line 138: I like the solution, but I wonder if we can keep the css out of the scope of the HTML to have a cleaner file. I thought about the header have a permanent margin bottom, and it wouldn't depend on the content displayed

vicarali: Solved! I've removed this style from this condition and set it up as a margin on the corresponding CSS file.

PR #234, opened 2022-07-06
Update/Insights-page-menu

RodrigoCastroS:
We removed the newsletter option from the Insights since it wasn't a benefit for the business and the team wasn't updating blogs with newsletter content


.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
vicarali, line 1: 😎

vicarali (approved):
LGTM!! 💯

PR #235, opened 2022-07-07
Update README.md

vicarali:
Add reference to the drag and drop sections found under the templates folder


RodrigoCastroS (approved):
Amazing! This makes me think that we could also move this kind of documentation in an extended format into Confluence. What do you think?

PR #236, opened 2022-07-07
Feature/case-study-template/headings-anchor

RodrigoCastroS:
Develop a new module mainly for the case studies consisting of a section divider with the option of using it as an anchor point


version_control/Codurance_September2020/modules/Section-divider.module/fields.json
```diff
@@ -0,0 +1,42 @@
...
+  "default" : false
+}, {
+  "id" : "572020da-37f2-93a6-3cdc-b1df37196ef5",
+  "name" : "anchor_id",
+  "display_width" : null,
+  "label" : "Anchor name",
+  "required" : false,
+  "locked" : false,
+  "validation_regex" : "",
+  "validation_error_message" : "The anchor name shouldn't contain whitespaces",
```
vicarali, line 30: Although I appreciate the idea to add this help message for this field, it seems that on the editor this validation does not occur even when adding blank spaces; I'm not sure if this type of validation can be configured here, but maybe just setting this exact message as an inline help text would be enough.

vicarali (approved):
Great job! I really like the approach of making the section dividers with a toggable anchor, not making it mandatory (and the icon of the module haha). 

I've simply left you a few comments regarding some minor details, if you want to discuss any of them just let me know!

version_control/Codurance_September2020/css/modules/section-divider.css
```diff
@@ -0,0 +1,19 @@
+
+{% import '../utils/utils.css' as utils %}
+
+.section-title {
+    display: flex;
+    gap: .8em;
+    align-items: baseline;
+
```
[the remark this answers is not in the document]
RodrigoCastroS: makes sense

version_control/Codurance_September2020/modules/Section-divider.module/fields.json
RodrigoCastroS: true, I forgot to remove it. I was trying to use the regex field but I couldn't make it work in the end

version_control/Codurance_September2020/modules/Section-divider.module/fields.json
```diff
@@ -0,0 +1,43 @@
...
+  "default" : false
+}, {
+  "id" : "572020da-37f2-93a6-3cdc-b1df37196ef5",
+  "name" : "anchor_id",
+  "display_width" : null,
+  "label" : "Anchor name",
+  "help_text" : "Write the anchor name with hyphens between words, otherwise, it won't work",
+  "required" : false,
+  "locked" : false,
+  "validation_regex" : "(?=\\S*[-])([a-zA-Z'-]+)",
```
vicarali, line 30: Damn, that's really impressive! To be fair I don't really understand how the first group of the regex works, with the S character in there. I tested it on online regex testers and it doesn't seem to work in the same way there, so I don't know if it's a Hubspot particularity, but in the latter works perfectly, so that's great!

If you are interested, I've figured out some small improvements that we could apply here: 
- I believe that the brackets are not required around the hyphen at the first group since it's a single character
- For the 2nd group, maybe we could remove the A-Z to disallow the usage of caps and the "'" character, since I believe it should not be used here

In case you want to test these changes out, please try the following string: "(?=\\S*-)([a-z-]+)"

vicarali (approved):
Great updates! I just left you one single comment regarding some possible improvements of the regex string, so please take a look if you are interested.

PR #237, opened 2022-07-08
Feature/case-study-template/cta-banner-theme-section

vicarali:
- Creation of a theme section for the Banner-Link module with default content for ease of use by the user
- Refactoring of Banner-Link module to remove the usage of the button macro (it didn't work when the module is used on a theme section and it wasn't mandatory)
- Fix the theme section of the testimonial slider (it wasn't working due to an incorrect path)
- Minor refactoring of Banner-Link and LP-testimonial-slider modules to include the CSS as module dependencies instead of via the require_css function
- Move the partials that previously where found on the "templates/partials/sections" folder to the parent folder "templates/partials" and delete the former folder. This was done to clarify the disctintion between partials and sections


RodrigoCastroS (approved):
LGTM! good work. 
I'm wondering that the fact of moving the author pages partial sections might not affect on the blog post. Also keep in mind you're changing that on the service line pages, so you might have to fill up the module with the content.

P.S. Merge master into your branch to check any issues

PR #239, opened 2022-07-12
Feature/case-study-template/text-and-image-theme-section

vicarali:
- Creation of new section template for the text-and-image module
- Add specific styles when the LP-testimonial-slider and text-and-image modules are used in dnd
- Set up new global CSS variable for the vertical spacing of dnd sections


version_control/Codurance_September2020/css/modules/lp-testimonial-slider.css
```diff
@@ -1,158 +1,170 @@
...
+/* Drag and drop styling */
+.dnd-section .testimonial-slider {
+  padding-bottom: 0;
 
-    .swiper-button-next:after,
-    .swiper-button-prev:after {
-        font-size:1rem;
+  /* Negative margin to allow the module to just use the padding set up       
+  globally on the dnd-section */
+  margin-top: calc(-1 * var(--dnd-vertical-spacing));
```
RodrigoCastroS, line 165: I don't know if having a negative margin to fix the padding top comming from the slider box might be the best solutiion. Maybe just chaching the .swiper-slide padding should work. Also bear in mind that this should work with several slides, so probably it might affect the styling as well

PR #241, opened 2022-07-19
Update/case-study-header

RodrigoCastroS:
We've changed the header:

* Background image and default icon
* New badge style and color

### DIFF
--- a/version_control/Codurance_September2020/css/pages/case-study-template-v2.css
+++ b/version_control/Codurance_September2020/css/pages/case-study-template-v2.css
@@ -9,37 +9,40 @@ body {
   max-width: unset;
 }
 
-/* Header Section */
+
 .case-study-header {
-  align-items: center;
-  background: url("https://info.codurance.com/hubfs/New%20Case%20Study%20Landing%20Pages%20Test/CS_header_bg.png");
-  background-position: top center;
-  background-size: cover;
+  background: url("https://3042464.fs1.hubspotusercontent-na1.net/hubfs/3042464/Case%20Study%20Landing%20Pages/case-study-LP-header-bg-img.svg");
+  background-position: top right;
+  background-size: 70%;
   background-repeat: no-repeat;
-  display: flex;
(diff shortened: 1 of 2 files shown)


version_control/Codurance_September2020/css/pages/case-study-template-v2.css
```diff
@@ -9,37 +9,40 @@ body {
...
-  background-size: cover;
+  background: url("https://3042464.fs1.hubspotusercontent-na1.net/hubfs/3042464/Case%20Study%20Landing%20Pages/case-study-LP-header-bg-img.svg");
+  background-position: top right;
+  background-size: 70%;
   background-repeat: no-repeat;
-  display: flex;
-  justify-content: center;
-  gap: 2em;
-  padding: 4em 2em 2em;
+  padding: 4em 2.5em;
```
vicarali, line 18: The padding-bottom for the header seems to be much smaller on the design from the extra_large to medium viewpoints. 

Other than that, I've also spotted that the padding-left on the medium viewpoint is a little bit bigger here than the rest of the sections of the page, but I don't know how we should approach that.

version_control/Codurance_September2020/css/pages/case-study-template-v2.css
```diff
@@ -62,44 +65,69 @@ body {
...
+    align-items: start;
+  }
+
   .case-study-header__text-block{
     max-width: none;
+    margin-top: unset;
   }
 
   .case-study-header__text-block .page-header.section-header h1 {
     {{ utils.freed() }}
```
vicarali, line 111: This font-weight seems to be incorrect on the small viewpoint; I believe we should add a font-weight: var(--heavy-font-weight); style here to reset the font-weight.

vicarali (approved):
The header looks spectacular now, very great job and solution! 

I've simply left you some comments regarding some minor details that maybe could be improved; please let me know if you want to discuss any of them.

PR #242, opened 2022-07-19
Feature/case-study-template/service-lines

vicarali:
- Creation of module to display a listing of Service Lines
- Creation of dnd section template to include the new module
- Minor fix: set the sizing type of the new header image to auto


version_control/Codurance_September2020/css/modules/service-lines-listing.css
```diff
@@ -0,0 +1,141 @@
...
+        {{ utils.natus() }}
+    }
+{% endcall %}
+
+.service-line__read-more-title,
+.service-line__read-more-close-button {
+    color: var(--mid-gray);
+    cursor: pointer;
+    font-weight: var(--heavy-font-weight);
+}
```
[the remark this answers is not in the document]
RodrigoCastroS: Just another addition for reference, the weight of the font is just bold, not extra bold

version_control/Codurance_September2020/css/modules/service-lines-listing.css
```diff
@@ -0,0 +1,141 @@
+{% import '../utils/utils.css' as utils %}
```
RodrigoCastroS, line 1: I have the feeling when reading the CSS file that I don't know your intentions when doing the layout. For example, it's really hard to figure out if you thought on building with mobile first in mind or desktop, since I can see that the media queries variates on the selector. It seems that you wrote as you were doing and what it was convinient instead having a first thougt on how it might be layed out.

version_control/Codurance_September2020/css/modules/service-lines-listing.css
```diff
@@ -0,0 +1,141 @@
...
+        flex-direction: column;
+    }
+{% endcall %}
+
+.service-line__image {
+    aspect-ratio: 1/1;
+    height: auto;
+    max-width: 100%;
+    min-height: 6rem;
+    min-width: 6rem;
```
RodrigoCastroS, line 30: Since you're using the aspect ratio, it seems that the min-height it's not necessary. Furthermore, I believe you won't need the min-width either, because the behaviour is already set, unless you want a specific width that in this case is the same as setting *width: 6rem;*

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/fields.json
```diff
@@ -0,0 +1,92 @@
...
+  "default" : [ {
+    "image" : {
+      "size_type" : "auto",
+      "src" : "",
+      "alt" : null,
+      "loading" : "lazy"
+    },
+    "read_more_button_text" : "Read more"
+  } ]
+} ]
```
RodrigoCastroS, line 92: Having a look at this and having used, I believe it makes more sense to have a group out of it. From the UX perspective it will look more intuitive when introducing the text and differenciating between the other fields

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.html
```diff
@@ -0,0 +1,36 @@
...
+
+            <p class="service-line__description {% if item.read_more_text %}service-line__description--read-more{% endif %}">
+                {{ item.description }}
+            </p>
+
+            {% if item.read_more_text %}
+                <details class="service-line__read-more">
+                    <summary class="service-line__read-more-title">{{ item.read_more_button_text }}</summary>
+                    <span class="service-line__read-more-text">{{ item.read_more_text }}</span>
+                    <button class="service-line__read-more-close-button">{{ item.read_more_button_text }}</button>
```
RodrigoCastroS, line 31: Since we're using a different html element to close the expanding text, having a discernible text explaining what it does it will help. For more reference, here: https://dequeuniversity.com/rules/axe/3.5/button-name or here: https://www.w3.org/WAI/standards-guidelines/act/rules/97a4e1/2022-01-28/

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
+
+const readMoreBlocks = document.querySelectorAll(".service-line__read-more");
+
+readMoreBlocks.forEach(
+    readMoreBlock => toggleReadMoreOnButtonClick(readMoreBlock)
+);
+
+function toggleReadMoreOnButtonClick(readMoreBlock) {
```
RodrigoCastroS, line 9: I'm not going to get into naming functions to deep, but it seems this is rather a really long name. For example, toggleReadMoreButton it might simplify. It explains what it does and to what.

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.js
```diff
@@ -0,0 +1,39 @@
...
+    const readMoreOpenButton = 
+        readMoreBlock.querySelector(".service-line__read-more-title");
+    const readMoreCloseButton = 
+        readMoreBlock.querySelector(".service-line__read-more-close-button");
+
+    readMoreOpenButton.addEventListener(
+        "click", 
+        () => { toggleReadMoreOverflow(readMoreBlock) }
+    );
```
RodrigoCastroS, line 18: I'm not a big fan of breaking up event listeners functions for readability when you can do it in one line and using arrow functions.

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.js
```diff
@@ -0,0 +1,39 @@
...
+
+const readMoreBlocks = document.querySelectorAll(".service-line__read-more");
+
+readMoreBlocks.forEach(
+    readMoreBlock => toggleReadMoreOnButtonClick(readMoreBlock)
+);
+
+function toggleReadMoreOnButtonClick(readMoreBlock) {
+    const readMoreOpenButton = 
+        readMoreBlock.querySelector(".service-line__read-more-title");
```
RodrigoCastroS, line 11: It seems that you're selecting a title, but naming it readMoreButton... seems a bit confusing to understand without looking at the behaviour at the web

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.js
```diff
@@ -0,0 +1,39 @@
...
+        () => { toggleReadMoreOverflow(readMoreBlock) }
+    );
+
+    readMoreCloseButton.addEventListener("click", collapseReadMore);
+}
+
+function collapseReadMore(event) {
+    const readMoreCloseButton = event.target;
+
+    readMoreCloseButton.parentNode.removeAttribute("open");
```
RodrigoCastroS, line 30: Maybe resolving this in one line? *event.target.parentNode.removeAttribute("open");*

RodrigoCastroS (commented):
Well done! :rocket
This one was a tricky one to complete but finally we finished the Case study template! (hopefully for some time). 
I've added a lot of comments (sorry for that), but using the module I've realised that the default text on the "read more" placeholder doesn't work unless you update the input which is a bit weird and could be confusing for future users.

Other than that, all good. If you want to discuss something else about comments I left, let me know.

version_control/Codurance_September2020/css/modules/service-lines-listing.css
vicarali: You are right! I set the aspect-ratio property after these ones and they became unnecessary. 

I've just changed the min-width property to width and removed the min-height one.

version_control/Codurance_September2020/css/modules/service-lines-listing.css
vicarali: As discussed, I've just refactored the code a little bit to follow a mobile-first approach, which I will continue doing moving forward.

version_control/Codurance_September2020/css/modules/service-lines-listing.css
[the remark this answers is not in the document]
vicarali: I've just corrected the sizing of the "Read more" button. 

Regarding the weight, as discussed it currently already uses bold (700) and not extra-bold, so we'll leave it as is.

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/fields.json
vicarali: Done!

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.html
vicarali: Done!

version_control/Codurance_September2020/modules/Service-Lines-Listing.module/module.js
vicarali: I've changed the class of the element to include the button keyword and now is clearer.

PR #243, opened 2022-07-25
Refactor conditional statement to show/hide button

RodrigoCastroS:
The optional "download case study" button wasn't appearing due to a misuse of a conditional.

It was fixed and addressed where the conditional should be.


vicarali (approved):
C'est magnifique! Just one small observation regarding the design that I've found: I've seen on the [design](https://xd.adobe.com/view/cd713f4d-684e-42c0-bdbb-4fda3d235447-9822/) that on the large/extra_large viewpoints there is a little bit more separation between the subtitle and the button, and that on mobile the width is more limited. I don't know if we should change any of those matters.

PR #244, opened 2022-07-28
Update/case-study-general-fixes

RodrigoCastroS:
Some changes were requested from the design team:

* The width of the secondary menu module
* New color and apply it to stats module
* Changes to the aspect ratio of the icons and gap on service lines moudule


vicarali (approved):
Good job!! The only adjustment I would do is the one we discussed earlier, change the padding of the secondary-menu on the large and extra_large viewpoints (it looks perfect on medium/small)

PR #247, opened 2022-08-08
Change toggle min width

RodrigoCastroS:
When the filter section has a long text description for it, the toggle seemed to be shrinking on mobile devices. We had to set up a min width to fix it.


vicarali (approved):
LGTM! 👏

PR #248, opened 2022-08-08
Refactor from flex to grid

RodrigoCastroS:


vicarali (approved):
Great job!! Before merging this I commited a small adjustment regarding the position:relative styles of some elements of the card items (we commented out by error one of them which was required and there was another one which was redundant).

PR #249, opened 2022-08-10
Change href url path

RodrigoCastroS:
As part of the upgrade on the videos page, we needed to change the link on the insights page menu


vicarali (approved):
LGTM!! 💯 🥇

PR #251, opened 2022-08-12
Block enter key on the search bar to prevent submit

vicarali:
The search bar filter on the videos page refreshed the page when the enter key was pressed on it, so we've blocked this key to avoid that

### DIFF
--- a/version_control/Codurance_September2020/modules/Events-Search-Bar.module/module.js
+++ b/version_control/Codurance_September2020/modules/Events-Search-Bar.module/module.js
@@ -4,8 +4,15 @@ const searchBarForm = document.querySelector(".events-search-bar");
 const searchBar = searchBarForm.querySelector(".events-search-bar__input");
 const searchBarResetButton = searchBarForm.querySelector(".events-search-bar__reset-button");
 
+searchBar.addEventListener("keypress", blockEnterKey);
 searchBar.addEventListener("input", filterEventsOnInputValueChange);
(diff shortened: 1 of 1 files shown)


version_control/Codurance_September2020/modules/Events-Search-Bar.module/module.js
```diff
@@ -4,9 +4,16 @@ const searchBarForm = document.querySelector(".events-search-bar");
 const searchBar = searchBarForm.querySelector(".events-search-bar__input");
 const searchBarResetButton = searchBarForm.querySelector(".events-search-bar__reset-button");
 
+searchBar.addEventListener("keypress", blockEnterKey);
 searchBar.addEventListener("input", filterEventsOnInputValueChange);
 searchBarResetButton.addEventListener("click", filterEventsOnResetButtonClick);
 
+function blockEnterKey(keypressEvent) {
```
RodrigoCastroS, line 11: I like the solution here, but being picky maybe I would choose dismissEnterKey() , because in reality you're not blocking it. Nothing too serious :)

RodrigoCastroS (approved):
I've left you a small comment related to naming the function, but it all works good! 👍

version_control/Codurance_September2020/modules/Events-Search-Bar.module/module.js
vicarali: Good idea! I've just changed it 👍

PR #252, opened 2022-08-15
Feature/kata index filter

jwilliamson15:


version_control/Codurance_September2020/modules/filtered-katas.module/module.html
```diff
@@ -0,0 +1,142 @@
+{{ require_css(get_asset_url("../../css/modules/filtered-clients.css")) }}
```
RodrigoCastroS, line 1: I wonder if we need to import all the styles from the filtered client page. Since we're using BEM convention to name our css classes it might be confusing landing on the Katas module and having something being referenced from the clients page.

version_control/Codurance_September2020/modules/filtered-katas.module/module.html
```diff
@@ -0,0 +1,142 @@
...
+{% import '../../partials/clients-page-cards.html' as cards %}
```
RodrigoCastroS, line 2: I believe you're not using this import, so probably it's dead code on this module

version_control/Codurance_September2020/modules/filtered-katas.module/module.html
```diff
@@ -0,0 +1,142 @@
...
+            </div>
+          {% endfor %}
+        </div>
+      </div>
+    </div>
+  </div>
+</main>
+
+{{ require_js(get_asset_url("../js/imageLoaded.js")) }}
+{{ require_js(get_asset_url("../../js/isotope.js"), "footer") }}
```
RodrigoCastroS, line 141: I haven't checked in depth this implementation, but since the filter it's a reusable file we had already we might not need this library. Although, we might want to keep the isotopePackery below which does all the animation when filtering

version_control/Codurance_September2020/modules/filtered-katas.module/module.html
```diff
@@ -0,0 +1,142 @@
...
+  topics: "Topics"
+}
+%}
+{% set infoMessage = { header: "We’re sorry, your search returned no results",
+paragraph: "Try adjusting your filters to find what you’re looking for."
+}
+%}
+
+
+<main class="clients-page__main">
```
RodrigoCastroS, line 26: As I said before, keeping other page naming for classes might be confusing

version_control/Codurance_September2020/templates/blog-index.html
```diff
@@ -79,12 +79,20 @@
                 Podcasts
               </a>
             </li>
+            <li class="custom-blog-listing__filter-item">
```
RodrigoCastroS, line 82: I know Laura mentioned to create an item element on the current blog index menu, but I'll probably avoid inserting it now because we have plans to include it on the main menu as a dropdown under Insights and also, when you'll merge the code you won't have a existing page to link to until you have the "katalyst blog" created.

RodrigoCastroS (changes requested):
Hi! I think overall it's good but I've left a few comments to review before merging. The comments are related to naming conventions, unused code and the item menu created. I hope it helps

version_control/Codurance_September2020/modules/filtered-katas.module/module.html
t-kellett: We've removed isotopePackery and not isotope (as removing isotope broke the animation, but removing Packery did not).

jwilliamson15:
Hi Rodrigo, we have made the changes suggested. Please take a look when you are free.

RodrigoCastroS (approved):
LGTM! Thanks for those changes, nice job!

PR #253, opened 2022-08-17
New page/faq template

RodrigoCastroS:
Create a simple FAQ template with existing elements to start populating it and maybe later improve on the design of it.


RodrigoCastroS (commented):
It seems all the changes won't have any conflict with any of the current modules on the site

PR #261, opened 2022-08-25
Remove current menu from Insights articles page

vicarali:
The new user journey will be defined on the new Insights overview page. 

In the interim, we'll add the current different links on the global navigation menu inside the Insights item.


RodrigoCastroS (approved):
All good! make sure to update master into this branch before merging, you might have conflicts

Well done @vicarali 🚀

PR #265, opened 2022-09-07
Update/insights-headers

RodrigoCastroS:
Create a header for pillar pages under the Insights tree


version_control/Codurance_September2020/css/partials/header-insights.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+{% call utils.medium_large_and_extra_large() %}
+    .header-insights__container {
+        flex-direction: row;
+    }
+   
+
+    .header-insights__image {
+        transform: scale(1.1) translate(10%, 5%);
```
vicarali, line 47: I belive that the transform property isn't really needed for this image on the large and extra_large viewpoints; on the design the max width set up for the header and the body of the pages is 1200px, without any overflow, so it shouldn't be required.

Feel free to tell me if you want to discuss this further.

version_control/Codurance_September2020/templates/blog-index.html
```diff
@@ -31,10 +31,11 @@
 </head>
 
 {% block body %}
-{% if blog_author %}
 
-  {% include "./partials/author-page.html" %}
+  {% include "./partials/header-insights.html" %}
```
vicarali, line 35: I've spotted a minor bug regarding this include: it seems to be loading also on author pages. I believe we should move this include at the top of the "else" block of the following conditional.

vicarali (approved):
Great job as always! :rocket I've just added two minor observations that I found, please feel free to tell me if you want to discuss anything.

version_control/Codurance_September2020/css/partials/header-insights.css
RodrigoCastroS: I know, I had a review with design already and we were pairing and they liked it as it is... even if it doesn't match exactly as the design

PR #268, opened 2022-09-15
Feature/video-page/filters

RodrigoCastroS:
New feature for the videos page, where you can filter by topic and language.


vicarali (approved):
I've added a commit cleaning some small parts of the changed files, like for example some consecutive media query utils calls that were redundant. Other than that, LGTM! 

I'm very proud of the results from the development of these new features!! Great work mate!

PR #269, opened 2022-09-22
Feature/blog-post/newsletter-popup-improvements

vicarali:
Blog posts improvements: Refactor pre-existing newsletter signup modules and implement feature to let Hubspot users to choose to display either a newsletter form or a CTA block at the end of the content.

- Hubspot users will be able to choose which module to show at the end of the content between the Newsletter-Form-Card (previously the Newsletter Inline Form) and the Blog Post Banner Recruitment modules
- Redesign Newsletter Footer Form module, removing the image and simplifying the language display method
- Simplify Blog Post Banner Recruitment module, removing unused features
- Augment the cookie from the Newsletter Form Popup module expiry date from 7 to 90 days

### DIFF
--- a/version_control/Codurance_September2020/css/newsletter-form.css
+++ b/version_control/Codurance_September2020/css/newsletter-form.css
@@ -82,7 +82,8 @@
     }
 }
 
-.hs-email input {
+.newsletter-footer-form__section .hs-email input,
+.newsletter-exit-popup__container .hs-email input {
     {{ utils.base() }}
     box-shadow: 0px 0px 20px #00000029;
     border: 1px white solid;
--- a/version_control/Codurance_September2020/css/pages/blog-post.css
+++ b/version_control/Codurance_September2020/css/pages/blog-post.css
@@ -170,6 +170,11 @@
   font-family: monospace;
 }
 
+.newsletter > .newsletter-wrapper {
+  margin-top: 2.4em;
+  margin-bottom: 2.4em;
+}
+
 {% call utils.extra_large() %}
   .newsletter > .newsletter-wrapper {
     max-width: var(--base-max-width);
@@ -186,6 +191,10 @@
 
   .newsletter__form {
     align-self: flex-end;
+  }
+
+  .newsletter .newsletter__form .hs-form {
+    align-items: unset;
   }
 {% endcall %}
 
--- a/version_control/Codurance_September2020/css/pages/blog-post.css
+++ b/version_control/Codurance_September2020/css/pages/blog-post.css
@@ -170,6 +170,25 @@
   font-family: monospace;
 }
 
+{% call utils.extra_large() %}
+  .newsletter > .newsletter-wrapper {
+    max-width: var(--base-max-width);
+  }
+
+  .newsletter-wrapper > .newsletter-inner {
+    flex-flow: column;
+    max-width: 30rem;
+  }
+
+  .newsletter-text-content__title {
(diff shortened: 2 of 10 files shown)


version_control/Codurance_September2020/css/pages/blog-post.css
```diff
@@ -170,6 +170,34 @@
   font-family: monospace;
 }
 
+.newsletter > .newsletter-wrapper {
```
RodrigoCastroS, line 173: I don't think we need the more than sing to find newsletter-wrapper class. The specificity will be enough to select the children from newsletter.

version_control/Codurance_September2020/css/newsletter-form.css
```diff
@@ -82,7 +82,8 @@
     }
 }
 
-.hs-email input {
+.newsletter-footer-form__section .hs-email input,
+.newsletter-exit-popup__container .hs-email input {
```
RodrigoCastroS, line 86: maybe I'll add also .hs-email input, because it might affect to the rest of the email inputs on the site, since the rules written seem a baseline

version_control/Codurance_September2020/templates/blog-post.html
```diff
@@ -185,10 +177,35 @@ <h2 class="blog-author-info__name">
...
+  {% if widget_data.CTA_or_newsletter_form.value == "CTA" %}
+    <section class="blog-recruitment-banner blog-max-width">
+      {% module "blog-recruitment-banner" path="../modules/Blog Post Banner Recruitment", label="CTA recruitment banner" no_wrapper=True %}
+    </section>
+  {% else %}
+    {% set newsletter_copy = config.newsletter_cta.en %}
 
+    {% if locale == "es" %}
+      {% set newsletter_copy = config.newsletter_cta.es %}
+    {% endif %}
```
RodrigoCastroS, line 197: Not a big fan of nested if statemnts, I believe we can solve this outside of the parent conditional? Although, it's a small improvement

RodrigoCastroS (approved):
Seems good to me, although the functionality works I left couple of comments related to CSS and a conditional. Great job! 🚀

version_control/Codurance_September2020/css/newsletter-form.css
vicarali: Good point! Done!

vicarali: I tried to re-add the selector that you just mentioned but it seems to break the styling from the Newsletter-Form-Card module (I believe that's the reason we did this change in the first place), so it seems that we will need to keep it out of here.

Even so, I believe it shouldn't affect negatively any field from the site, because this CSS file is only included in the three newsletter modules from the blog (Newsletter Inline Form, Newsletter Exit Popup, Newsletter Footer Form)

version_control/Codurance_September2020/css/pages/blog-post.css
vicarali: You're right, I just corrected it on this line and some selectors from the following media query

version_control/Codurance_September2020/templates/blog-post.html
vicarali: Makes sense; I just moved this block at the top of the template, just below the import of the config.html file, which seems to be a better location.

I've also seized the opportunity to simplify a little bit the set up of language for the Exit Popup module by using the locale object, because it was kind of verbose before.

PR #272, opened 2022-10-03
Add fading animations to the Katalyst listing cards

vicarali:
The kata cards from Katalyst didn't have any animations before, so we've added a fading effect.

### DIFF
--- a/version_control/Codurance_September2020/modules/filtered-katas.module/module.js
+++ b/version_control/Codurance_September2020/modules/filtered-katas.module/module.js
@@ -251,8 +251,8 @@ const refilter = _ => {
   katas.visible = calculateVisibleKatas();
   katas.hidden = arrayDifference(katas.all, katas.visible);
 
-  katas.visible.forEach(show);
-  katas.hidden.forEach(hide);
+  katas.visible.forEach(showWithAnimation);
+  katas.hidden.forEach(hideWithAnimation);
 }
 
 const update = _ => {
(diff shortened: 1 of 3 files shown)


version_control/Codurance_September2020/modules/filtered-katas.module/module.js
```diff
@@ -325,11 +325,43 @@ const addRemoveFilterListeners = _ => {
...
+}
+
+function hideWithAnimation(element) {
+  addFadeAnimationModifier(element);
+      
+  // Timeout to show the transition before the display property changes
+  setTimeout(addHiddenModifier, getFadingAnimationDuration(), element); 
+}
+
+function getFadingAnimationDuration() {
```
RodrigoCastroS, line 343: I've been thinking about this function, and we might not need it. Because we basically we're getting the animation duration and returning it as an Integer. Maybe retrieving the this info on hide with animation, I don't think it would be an issue reading the functionality.
For example doing "const animationDuration = 
      getComputedStyle(document.documentElement).getPropertyValue("--fading-animation-duration");"

version_control/Codurance_September2020/modules/filtered-katas.module/module.js
```diff
@@ -325,11 +325,43 @@ const addRemoveFilterListeners = _ => {
...
+  const rootHtml = document.documentElement;
+  const fadingAnimationDurationInMs = 
+      getComputedStyle(rootHtml).getPropertyValue("--fading-animation-duration");
+  
+  return parseInt(fadingAnimationDurationInMs);
+}
+
+function removeFadeAnimationModifier(element) {
```
RodrigoCastroS, line 351: I'm wondering if these functions can be transformed into toggle, since they're only adding and removing the same class element.

RodrigoCastroS (approved):
It seems it's working on my end. I left a couple of suggestions that could be followed if you want.
Maybe, to think as an improvement in the future, the cards could dissapear in a more gentle way and less sloppy, so the choreography is better implemented.

version_control/Codurance_September2020/modules/filtered-katas.module/module.js
vicarali: Alright! If you find it readable enough I'll do just that and then we can get rid of one more function

version_control/Codurance_September2020/modules/filtered-katas.module/module.js
vicarali: As discussed, we won't change this functionality for now because currently the filtering function seems to be executed several times at once.

PR #275, opened 2022-10-05
Enhancement/upcoming-events-language

RodrigoCastroS:
Until now, we had mixed languages on the events page. We decided to move into showing the content depending on the domain name. On the English site there will be only events hold in that language, on the other hand, if the site is in Spanish the user will see all events in both languages.

To make this work, we had to create a language column on the upcoming events DDBB table.


vicarali (approved):
Tested and approved! Good job!

PR #276, opened 2022-10-11
Indent ul ol lists and style numeric lists on posts

RodrigoCastroS:
Create the option for users writing in the blog post, to use ordered lists.

In addition, I've indented ordered and unordered lists, so the bullets are being within the content width.


vicarali (approved):
Good job! I've simply added a few minor commits to correct a browser incompatibility and to remove the currently deprecated kata-post template.

The removal of the template was required because the blog-content class is specific to the general blog-post template, and since we prepared this latter template to be used also on the Kata Posts from now on I thought it was a good moment to do so.

PR #278, opened 2022-10-27
Feature/blog-search

RodrigoCastroS:
On this feature we had:

- Create a blog search on our publications page, where the user can be redirected to a page where all the relevant results are shown.

- The displayed grid of the latests post was improved from previous stages. We removed dead code and simplified css selectors.

- Adapt the insights headers for laptop screens so the above the fold includes the top of the cards displayed.


vicarali (approved):
Good job mate! I'm really happy about the results of this feature!

I've simply added a few commits to adjust some minor details that I found here and there, but everything's in a really good state overall.

PR #282, opened 2022-11-16
New-template/rainbow-roles-page

vicarali:
- Develop new template for the Rainbow Roles page

- Creation of two new modules: Clickable-Listing to set up a list of cliclable image links and Iframe-Listing to set up a list of iframe elements

- Remove unused files inside the _locales folder from the Locations module


version_control/Codurance_September2020/css/modules/locations.css
```diff
@@ -53,8 +53,16 @@
...
 
-.location__img{
+.location__img {
     width: 10em;
+    clip-path: circle();
+    transition: transform ease-in-out 200ms;
+    
+}
+
+.location__img:hover {
```
RodrigoCastroS, line 63: I wonder why this appears here if you already deleted it... maybe is a dubplicated?

version_control/Codurance_September2020/css/pages/rainbow-roles.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+{% call utils.large() %}
+  html {
+    scroll-padding-top: 4em;
+  }
+{% endcall %}
+
+{% call utils.large_and_extra_large() %}
+  .header-hero > .header-hero__container {
```
RodrigoCastroS, line 15: I don't think you need the ">" character here, since you only have one container is the only one that is going to find. I would recommend this when you have a listing with the same class ;)

RodrigoCastroS (approved):
Seems good to me. Good work! 🚀

version_control/Codurance_September2020/css/modules/locations.css
vicarali: That's because this is from the old Locations module, not the new Clickable-Listing one haha, so I guess it's fine to leave it there

version_control/Codurance_September2020/css/pages/rainbow-roles.css
vicarali: I had to do it like that because if not it didn't seem to have enough specificity to be able to override these styles

PR #284, opened 2022-11-18
Enable translation for options of language filter

vicarali:


RodrigoCastroS (approved):
LGTM, although we might have in mind having just the languages in Spanish since we won't use them on the EN site

PR #286, opened 2022-11-23
New template/insights overview page

ndowkunda:
## Template Contains;

# Partial html:

- header-insights -> Editable title and subtitle

# Modules:

- recent-posts-v2 -> Latest three blog posts
- editors-pick  -> CMS user can select a maximum of 2 blog posts (w/thumbnails) and 4 pages from the site
- banner-link -> CTA section banner
- codurance-in-motion -> CMS users have the choice to choose to display between 1 & 5 videos from our existing Hubspot DB, videos played in page. 

# Assumptions: editors-pick module -> blog posts and pages are all selected via a **one** page field (with repeater)


version_control/Codurance_September2020/templates/insights-overview-page.html
```diff
@@ -0,0 +1,34 @@
...
+    label="Banner Link" no_wrapper=True %}
+  </section>
+  <section class="insights-overview-videos">
+    {% module 'codurance-in-motion' path='../modules/Codurance-in-motion.module'
+       label="Codurance in motion" no_wrapper=True %}
+  </section>
+</main>
+{% endblock body %}
+
+{% block styles %}
```
RodrigoCastroS, line 32: Probably I would move this to the top of the file. And I believe there isn't necessary to add a block style to add a require. I'm guessing this was created for a reason that is different from the other pages

version_control/Codurance_September2020/modules/Editors-Pick.module/module.html
```diff
@@ -0,0 +1,66 @@
...
+    {% endif %}
+
+    <div class="cm-editors-pick__items">
+      {% set blog_posts = module.page_field %}
+      {% for content_id in blog_posts %}
+        {% set size = loop.index <= 2 ? "big" : "small" %}
+        {% set blog_post = content_by_id(content_id) %}
+        <div class="cm-editors-pick__item">
+          {% if loop.index <= 2 %}
+            <div class="cm-editors-pick__image">
```
RodrigoCastroS, line 20: I wonder if there's a way of avoid excessive nesting tags with divs and using more relevant tags such <article> or <h3> for the featured title.

version_control/Codurance_September2020/css/utils/_typography-utils.css
```diff
@@ -74,6 +74,12 @@
   line-height: var(--natus-line-height);
 {% endmacro %}
 
+{% macro natus_reduced() %}
```
RodrigoCastroS, line 77: I don't think there's a need to create this macro. We can discuss this further so I can explain

version_control/Codurance_September2020/css/modules/codurance-in-motion.css
```diff
@@ -0,0 +1,61 @@
...
+
+.cm-codurance-in-motion__summary {
+
+}
+
+.cm-codurance-in-motion__cta {
+    margin-top: 2rem;
+}
+
+.cm-codurance-in-motion__heading-container {
```
RodrigoCastroS, line 31: I would remove unused CSS selectors

version_control/Codurance_September2020/css/modules/codurance-in-motion.css
```diff
@@ -0,0 +1,61 @@
...
+    max-width: var(--content-max-width);
+    margin: 0px auto;
+    padding-bottom: 1rem;
+    display: grid;
+    column-gap: 1rem;
+    row-gap: 1rem;
+}
+
+.cm-codurance-in-motion__heading {
+    {{ utils.eames() }}
```
RodrigoCastroS, line 18: This seems a bit counter intuitive. Since the size of the text is dudler, Why using eames utils?

version_control/Codurance_September2020/css/modules/recent-posts-v2.css
```diff
@@ -1,11 +1,9 @@
 {% import '../utils/utils.css' as utils %}
 
-
 .cm-recent-post {
   max-width: var(--content-max-width);
   margin: 0 auto;
-  transform: translateY(-20%);
-
+  transform: translateY(-5rem);
```
RodrigoCastroS, line 6: Careful with this because this module is used in some other places, so this change might affect other layouts. Please, double check where this might occur. You can overwrite this rule on the page template css

version_control/Codurance_September2020/css/templates/insights-overview.css
```diff
@@ -0,0 +1,17 @@
+{% import '../utils/utils.css' as utils %}
+
+.insights-overview-cta {
+    padding: 1rem;
+    color: #FFF;
+    {{ utils.color1_to_color2('271deg', '#1D233C', '#0C0C0D)') }}
```
RodrigoCastroS, line 6: I believe degrees won't work in here. You could use "to top", "to bottom" directives because the gradient is not beign applied. Or you could use the macro "slate_to_woodsmoke"

RodrigoCastroS:
I've created some comments on some of the files. If you think there's something that isn't clear let me know and we can talk about it :)

version_control/Codurance_September2020/css/templates/insights-overview.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.insights-overview-cta .banner {
```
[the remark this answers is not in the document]
antony-scott: This rule is nested & very specific in order to override the behaviour of the `Banner-Link` module and to avoid having to expose another field to control that behaviour. I've checked the behaviour and it didn't need the `!important` tag.

version_control/Codurance_September2020/css/pages/insights-overview-page.css
```diff
@@ -0,0 +1 @@
+{% import '../utils/utils.css' as utils %}
```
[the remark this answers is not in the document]
antony-scott: good spot, thanks - file will be removed in upcoming commit

version_control/Codurance_September2020/templates/insights-overview-page.html
antony-scott: you're right, done in upcoming commit

version_control/Codurance_September2020/modules/Editors-Pick.module/module.html
antony-scott: agreed, we've changed it to use `<section>`, `<article>`, `<figure>`, `<h3>` and `<p>` elements

version_control/Codurance_September2020/css/modules/recent-posts-v2.css
ndowkunda: Double checked now and confirmed `transform: translateY(-5rem);` does shift the `recent-posts-v2` module down by approx. 1cm or so in `home-v2.html`, this has been changed as suggested in upcoming commit 👍

version_control/Codurance_September2020/css/utils/_typography-utils.css
RodrigoCastroS: This macro definition won't be needed since is not in use.

version_control/Codurance_September2020/css/base/_typography-2021.css
```diff
@@ -36,6 +36,8 @@
   --base-margin: 1.3809524rem;
   --base-max-width: 65ch;
 
+  --reduced-font-size-multiplier: 0.79;
```
RodrigoCastroS, line 39: Since this variable is not in use, it might be best to remove it

version_control/Codurance_September2020/css/modules/editors-pick.css
```diff
@@ -0,0 +1,131 @@
...
+    max-width: calc(var(--content-max-width) + 4rem);
+    margin: 0px auto;
+    padding-inline: 2rem;
+}
+
+.cm-editors-pick__items {
+    margin-bottom: 1.5rem;
+}
+
+.cm-editors-pick__items, .cm-editors-pick__big_items, .cm-editors-pick__small_items {
```
RodrigoCastroS, line 19: When reading the code it seems the flex property is beign redefined on every breakpoint. Maybe the best approach would be thinking what's the first approach with the layout: is it Desktop? is it Mobile? And from there build up the layout, without the need of using breakpoint here and there.

version_control/Codurance_September2020/css/partials/header-insights.css
```diff
@@ -10,7 +10,8 @@
 .header-insights__container {
     display: flex;
     flex-direction: column-reverse;
-    max-width: var(--content-max-width);
+    max-width: calc(var(--content-max-width) + 4rem);
```
RodrigoCastroS, line 13: I'm wondering if this calc and padding for the container is necessary. Bear in mind that this header is used accross other 4 pages.

version_control/Codurance_September2020/css/templates/insights-overview.css
```diff
@@ -0,0 +1,110 @@
+{% import '../utils/utils.css' as utils %}
+
+.recent-blog-posts {
+    margin-bottom: -3rem;
+}
+
+{% call utils.medium() %}
+    .insights-overview-hero .header-insights, .recent-blog-posts .cm-recent-post, .editors-pick .cm-editors-pick__container, .insights-overview-cta .banner, .insights-overview-videos .cm-codurance-in-motion__container {
```
RodrigoCastroS, line 8: There's a helper class for lateral paddings called lateral-spacing, that could be really useful for this instead of applying the same rule to many classes. Also the readability would be improved.

version_control/Codurance_September2020/modules/Editors-Pick.module/module.html
```diff
@@ -0,0 +1,81 @@
...
+<div class="cm-editors-pick__container">
+  {% if module.heading %}
+    <h2 class="cm-editors-pick__heading">{{ module.heading }}</h2>
+  {% endif %}
+
+  {% set blog_posts = module.page_field %}
+
+  <section class="cm-editors-pick__items">
+    <div class="cm-editors-pick__big_items">
+      {% for content_id in blog_posts %}
```
RodrigoCastroS, line 16: Victor and I were wonder if there would be a way to simplify the loops created. For example, saving the blog posts in 2 different arrays: one with the featured, another with the small ones. In this case you'll avoid iterating the same array twice and also the conditional regarding the loop index values.

antony-scott: I prefer having the posts/pages in one line with the first 2 being "featured". We can move to 2 separate lists if that's preferable, but IMO the UI/UX in the page editor would then become quite cumbersome. I spent a while trying to make these 2 loops as clean as possible, but the HUBL templating language is quite limiting.

version_control/Codurance_September2020/css/partials/header-insights.css
antony-scott: I've moved the padding & width rules to insights overview specific css. The max width `calc` is required to keep the heading text indent in line with the rest of the insights overview page.

version_control/Codurance_September2020/modules/Editors-Pick.module/module.html
vicarali: Hello Antony,  we were only talking about the current "blog_posts" array, which maybe could be splitted to two different arrays, one for each type of post. 

Doing that, it wouldn't be required to modify the CMS fields of the module.

RodrigoCastroS (approved):
LGTM to merge. I've made a commit to fix an issue with the width of the header insights.

PR #287, opened 2022-11-23
Update/serverless-keys

RodrigoCastroS:
Update API token keys for the newsletter popup event that is triggered when it's a new user not registered in our database.


vicarali (approved):
LGTM! Good job, I'm glad we finally got that out of the way 👏

PR #294, opened 2022-12-14
Update/add-subtitle-to-video

RodrigoCastroS:
Add a feature of including the role title and person name under the videoAsk iframe. 

We also added a scroll into view when clicking any of the circles.


vicarali (approved):
LGTM 🤖

PR #295, opened 2022-12-15
Insights overview hotfix

antony-scott:
Hot fixes for the following issues ...

- editor's pick column sizing inconsistent (changed to use css-grid, instead of flex-box)
- editor's pick featured post image sizes differ (as above)
- title using internal name for pages (change to use title instead of name for pages)
- missing subtitle for pages (changed to use meta_description rather than first 150 characters of postBody)
- broken image link appearing when page/blog post has no featured image
- recent-posts-v2 - when no featured image, text appears at top, which results in misalignment of text when there is a mix of recents posts some with featured images, some without (use tango-to-white gradient instead of image)

### DIFF
--- a/version_control/Codurance_September2020/css/modules/editors-pick.css
+++ b/version_control/Codurance_September2020/css/modules/editors-pick.css
@@ -13,22 +13,24 @@
 }
 
 .cm-editors-pick__items {
-    display: flex;
-    flex-direction: row;
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr;
     gap: 1.5rem;
     margin-bottom: 1.5rem;
 }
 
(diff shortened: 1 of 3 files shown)


version_control/Codurance_September2020/css/modules/editors-pick.css
```diff
@@ -13,22 +13,24 @@
 }
 
 .cm-editors-pick__items {
-    display: flex;
-    flex-direction: row;
+    display: grid;
```
vicarali, line 16: I like the new approach with the grid a lot! Just a minor observation: if you defined the grid to work only on the medium_large_and_extra_large viewpoint, you shouldn't even need to set up any related styles on the small viewpoint since it's just 1 column

PR #298, opened 2022-12-21
Insights-overview/firesidechat

RodrigoCastroS:
- Implemented a section with the latest 2 firesidechats. If the user wants to select preferred ones, they can do it as well.

- Also, we decided to clean up a deprecated method from Hubspot related to the date

- Deleted an unused module called Services Media


vicarali (approved):
LGTM! 🤖

PR #303, opened 2023-01-09
Update/insights-overview-page/katalyst-section

vicarali:
Create new section on the Insights Overview page to showcase kata-related content with the following features:

- Row showing randomised katas from Katalyst
- Row displaying latest recorded Screenkatas


RodrigoCastroS (approved):
Looks good to me. Have a look at random katas module to rename the css classes, it might be an improvement.

PR #304, opened 2023-01-09
Dependencies/remove-fontawesome

RodrigoCastroS:
Delete Font Awesome library and its depencencies since we are working now with Line Awesome. Also, the head of the blog was removed since it was redundant and coming from the base html file


version_control/Codurance_September2020/templates/partials/author-page.html
```diff
@@ -29,31 +29,31 @@ <h1 class="blog-author__title">{{ blog_author.display_name }}</h1>
...
+                    <i class="lab fa-twitter"></i> 
                     </a>
                 </li>
                 {% endif %}
                 {% if blog_author.website %}
                 <li class="list-inline-item custom-mx-2">
                     <a class="blog-social-icons__icon" 
                     href="{{ blog_author.website }}" target="_blank" rel="noopener noreferrer"> 
-                    <i class="fa fa-link"></i> 
+                    <i class="lab fa-link"></i>
```
vicarali, line 40: These logos don't seem to load correctly. I think that to use Line Awesome here we have to change the fa- to la- on the specific classes for each logo.

Also, the logos seem to be a little bit smaller than the previous ones, maybe we could add a font-size style to correct that.

vicarali (approved):
Good job!! I simply added a single comment regarding an issue I've found on the author page social logos. Other than that, LGTM!

PR #307, opened 2023-01-11
Update/insights/blog-search-system

vicarali:
- Add date and author to the search results page from the Insights Blog
- Adjust max-width from the search results listing
- Refactor search bar JS to use arrow functions and improve readability


RodrigoCastroS (approved):
LGTM! great team-work 🚀

PR #309, opened 2023-01-13
Set default margin on buttons only on site pages

RodrigoCastroS:


version_control/Codurance_September2020/css/modules/button.css
```diff
@@ -1,11 +1,11 @@
 {% import '../utils/utils.css' as utils %}
 
-.custom-btn {
+.hs-site-page .custom-btn {
   margin-top: 2em;
 }
 
 {% call utils.small() %}
-  .custom-btn .button-primary {
+  .hs-site-page .custom-btn .button-primary {
```
vicarali, line 8: I'm not sure if we would need to add the class also to this selector, I believe these width styles would be nice to have globally and shouldn't be a problem 🤔

vicarali (approved):
LGTM! 👏 I've simply left a comment regarding a minor detail

PR #311, opened 2023-01-18
Update/isights/upcoming-events

RodrigoCastroS:
Create a section displaying the upcoming events using simple cards with limited information.

### DIFF
--- a/version_control/Codurance_September2020/css/modules/simple-card.css
+++ b/version_control/Codurance_September2020/css/modules/simple-card.css
@@ -36,8 +36,14 @@
     height: 100%;
     display: flex;
     flex-direction: column; 
-    justify-content: space-between;
 }
+
+.card__info time {
+    {{ utils.sejima() }}
+    margin-bottom: .5em;
+    text-transform: uppercase;
+}
+
 
 .card__info h3 {
     {{ utils.freed() }}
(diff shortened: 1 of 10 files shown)


version_control/Codurance_September2020/css/modules/simple-card.css
```diff
@@ -0,0 +1,62 @@
+{% import '../utils/utils.css' as utils %}
+    .card-container {
+        display: grid; 
+        grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
+        gap: 1em;
+    }
+
+{% call utils.large_and_extra_large() %}
+    .card-container {
+        grid-template-columns: repeat;
```
vicarali, line 10: The grid-template-columns: repeat here doesn't seem to work. Even so, I just tried all the styles from this selector on large and extra_large and other than the gap they don't seem to be changing anything, so I think we could get rid of these 3 first styles.

vicarali (approved):
Good job mate! Another great section finished 🚀 

I just left a single comment regarding a few CSS styles that perhaps we could remove, but other than that everything's perfect 👏

version_control/Codurance_September2020/css/modules/simple-card.css
RodrigoCastroS: Yes, you're right! indeed there's some rules that I have to take out. With the repeat auto-fit it already works as intended

PR #312, opened 2023-01-18
Display  events shown on the EN and ES site

RodrigoCastroS:
There was an issue related to the events that were displaying. Now on the EN site only English events will show. On the ES site, both languages will appear.


vicarali (approved):
LGTM 🤖

PR #319, opened 2023-02-06
Change sorting label on the Case Study cards group

vicarali:
Previously, the group from the Case Study Card Collection module was being sorted by the alt text from the image. Now it will be sorted by the card title.


RodrigoCastroS (approved):
Looks good for me 👍🏼

PR #320, opened 2023-02-07
Fix/newsletter-banner-layout

RodrigoCastroS:
We found an issue with the layout on the newsletter banner module. There was an overflow on the screens.

Also we improved:
* The markup for the module
* The aria-label where the module was being used


vicarali (approved):
LGTM! 🏷️

PR #321, opened 2023-02-14
Update/footer/office-addresses

RodrigoCastroS:
We had a long list of addresses in our footer and decided to slim down the content displayed.

Now it will display:
* General codurance e-mail address
* The offices cities, when clicked in one will redirect to the contac us page to the right location and contact info


vicarali (approved):
LGTM! Good job 🥇

PR #323, opened 2023-02-14
Feature/blog-posts/custom-cta-banner

vicarali:
Enable option to set up a custom CTA banner on blog posts. 

Its current intended use is for new CTAs specific to the ES site.


RodrigoCastroS (approved):
Looks good to me! 🦖

Issue #325, opened 2023-02-15
Youtube embed videos not rendering properly 


RodrigoCastroS:
There's some issue about the videos of the firesidechat not rendering. It seems is something related on how we strip down the url to get the youtube id.

Here's the example: 
`{% if video.video is string_containing "youtube" %}
			{% set youtube_video_id = video.video | split("v=") | last | truncate(11, True, "") %}
		{% else %}
			{% set youtube_video_id = video.video | split("/") | last %}
		{% endif %}`

PR #326, opened 2023-02-15
Fix how the youtube tag was beign recovered

RodrigoCastroS:
Resolves: #325


vicarali (approved):
LGTM! Good job mate! 🤖

PR #328, opened 2023-02-17
Update/service-lines/what-you-gain-from-working-with-us

vicarali:
Adjustments related to the what you gain from working with us section on the Service Line pages to make it more concise:
- Create field on Text-and-image module to be able to align the image vertically
- Style adjustments on the section of the template and the aforementioned module


RodrigoCastroS (approved):
LGTM! 👍🏼

PR #329, opened 2023-02-17
Style: remove unnecesary padding on dnd sections

RodrigoCastroS:


vicarali (approved):
LGTM! Very happy to get rid of these paddings 😁

PR #331, opened 2023-02-24
Update/replace-recent-posts-module

RodrigoCastroS:
With the use of different modules we found out that we could use just one, so we refactored some code to use one module and the delete the old/unused one


vicarali (approved):
LGTM! 💯

PR #335, opened 2023-03-02
Update/replace-module/community-events-listing

vicarali:
Refactoring of the previous event listing modules (Upcoming Events Listing and Recent-Events-Listing) in order to remove the Community Events Listing, which will be replaced with an additional instance of the Upcoming Events Listing.

### DIFF
--- a/version_control/Codurance_September2020/js/cardSlider.js
+++ b/version_control/Codurance_September2020/js/cardSlider.js
@@ -7,12 +7,7 @@ class CardSlider {
     cardsSelector = "[data-cardslider-card]",
     leftButtonSelector = "[data-cardslider-button-left]",
     rightButtonSelector = "[data-cardslider-button-right]",
-    filterButtonSelector = "[data-card-filter-button]",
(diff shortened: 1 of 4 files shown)


version_control/Codurance_September2020/js/cardSlider.js
```diff
@@ -2,51 +2,37 @@ class CardSlider {
   constructor({
     activationPoint = 1023,
     navigationControl = true,
-    cardWindowSelector = "[data-cardslider-window]",
+    slider = document.querySelector("[data-past-card-window]"),
```
RodrigoCastroS, line 5: I'm realizing now that we could have a more generic name for this. Because, if we place [data-past...] or [data-recent...] are really attached to a certain to a specific "action". So maybe a generic one would work, and we can override it passing it through the constructor

RodrigoCastroS (approved):
LGTM 🚀 Although I left you a little comment on setting the default values for one of the parameters.

PR #336, opened 2023-03-08
Fix/our-people-loading-assets

RodrigoCastroS:
We had currently a long listing of people working for us, thus performance issues came along due different image sizes.

I've added:
* A srcset specifically for mobile to improve loading times and LCP
* A resizing hubspot method to set a default value without thinking on how users upload their images
* A set width and height, so we can lazy load images

### DIFF
--- a/version_control/Codurance_September2020/modules/Our-People-Listing.module/module.html
+++ b/version_control/Codurance_September2020/modules/Our-People-Listing.module/module.html
@@ -20,7 +20,12 @@
         <div class="custom-our-people__card-inner" data-modal-target="#team-{{ loop.index }}">
           {% if row.image.url %}
             <img class="custom-our-people__profile-image" 
-              src="{{ resize_image_url(row.image.url, 300) }}"
+              srcset="
+                {{ resize_image_url(row.image.url, 180) }} 180w,
+                {{ resize_image_url(row.image.url, 300) }} 500w"
+              sizes="
+                (max-width: 400px) 180px,
+                500px"
               width="300" height="300" loading="lazy" 
               alt="{{ row.name }}" >
           {% endif %}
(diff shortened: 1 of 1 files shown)


version_control/Codurance_September2020/modules/Our-People-Listing.module/module.html
```diff
@@ -19,7 +19,15 @@
       <article class="custom-our-people__card  {% if row.leadership_board %}leadership{% endif %}">
         <div class="custom-our-people__card-inner" data-modal-target="#team-{{ loop.index }}">
           {% if row.image.url %}
-            <img class="custom-our-people__profile-image" src="{{ row.image.url }}" alt="{{ row.name }}">
+            <img class="custom-our-people__profile-image" 
+              srcset="
+                {{ resize_image_url(row.image.url, 180) }} 180w,
+                {{ resize_image_url(row.image.url, 300) }} 500w"
```
vicarali, line 25: Is there any specific reason for using the 500 value on this line and the 28th? 

If that's not the case, I believe it would be better to set it as 300, since it is the size defined for the image on this resize function.

version_control/Codurance_September2020/modules/Our-People-Listing.module/module.html
```diff
@@ -19,7 +19,15 @@
...
+              sizes="
+                (max-width: 400px) 180px,
```
vicarali, line 27: Since some mobile screens arrive to almost 425px, maybe we could increase the max-width to that and the size a little bit (200px~)

vicarali (approved):
Good job! I left you a few comments regarding some minor possible improvements, but LGTM!

version_control/Codurance_September2020/modules/Our-People-Listing.module/module.html
RodrigoCastroS: changed!

PR #341, opened 2023-03-17
Style: create a default spacing for mobile and tablets

RodrigoCastroS:
We found that by default there wasn't a margin on small devices on the case study template page.


vicarali (approved):
LGMT! Great solution! 🥇

PR #343, opened 2023-03-23
Create/industries/individual-pages-template

vicarali:
- Create template for the individual Industry pages
- Style Hubspot menus of breadcrumb type
- Adjust spacing and sizing of several modules used on the template to make them more consistent
- Creation of a CSS helper class named "region" to set up default vertical paddings on sections
- Set up a utils macro to define a global border radius for all card-like components
- Add option on Text-and-image module to hide the image on mobile view


RodrigoCastroS (approved):
Looks good to me to send it to production. What I've seen is that we could define a better strategy on where to write the breakpoints so we keep consistency all over the CSS files.
Let's keep up the good work, we are making nice improvements! :)

PR #344, opened 2023-03-23
Configure landing pages to have default padding

vicarali:
The dnd sections were forced to not have padding, which we want to have.


RodrigoCastroS (approved):
LGMT 🚀

PR #345, opened 2023-03-28
Update/portuguese-translations

RodrigoCastroS:
There was a request from the Portugal team to translate some of the fields that we had on our clients filters and current job listing. For more info, [refer to the request](https://codurance-unit.monday.com/boards/1582824572/pulses/4199240677)


vicarali (approved):
LGTM! Thank you for handling these PT translation issues 👏

PR #357, opened 2023-04-18
Create/industries/overview

RodrigoCastroS:
As a user I want to have an overview page to see all the industries Codurace has worked with. We've created new and reused some modules to build this page.


vicarali (approved):
LGTM! I commited a few code cleaning adjustments regarding several CSS styles and HTML data attributes that weren't being used but other than that everything's great!!

PR #361, opened 2023-04-19
Fix/industries-header

RodrigoCastroS:
Updade with last minute changes the font sizing for the header on the industries overview


vicarali (approved):
LGTM!

PR #363, opened 2023-04-20
Update/module/text-and-image-spacing-config

vicarali:
Remove as many typography, spacing, sizing and gap page-specific styles from Text-and-image modules in order to unify them and improve consistency.


version_control/Codurance_September2020/css/modules/text-and-image.css
RodrigoCastroS, line 1: Maybe deleting this block of code was the initial idea of the spacing issues we were having, right? 
`.text-and-image__image img {
    transform: translate(20%);
}
`

version_control/Codurance_September2020/css/pages/careers-v2.css
```diff
@@ -91,62 +91,21 @@
...
     {{ utils.black_to_navy() }}
     color: white;
     padding: 2em 0;
 }
 
-.why-join-us .text-and-image__copy p > a{
-    color: var(--tango-400);
-}
-
-.why-join-us .text-and-image__copy ol{
```
RodrigoCastroS, line 106: By deleting this, we'll miss the enumeration on the list

version_control/Codurance_September2020/css/pages/careers-v2.css
```diff
@@ -91,62 +91,21 @@
...
-    background: white;
-    color: var(--persimmon);
-}
-
-.why-join-us .text-and-image__image {
-    scale: 1.2;
-    align-self: start;
-}
-
-.why-join-us .container--img-left img {
```
RodrigoCastroS, line 125: I think this translation was on purpose on large screens. Mainly when reaching tablet breakpoints

RodrigoCastroS (changes requested):
I've left a couple of comments and I'll add a commit, but there are some changes you made we might want to talk about because it also will impact some pages we currently have.

Also it's worth having a list of those pages where we hide the image from the module, so we apply that functionality on the production environment.

RodrigoCastroS (approved):
All good! What an improvement! Happy that we are doing small changes that are helping us on our development workflow.

PR #372, opened 2023-04-26
Changes from PT review

tfbica:


RodrigoCastroS (approved):
Seems good to me. I've added a couple of commits changing the wording for the Spanish text and also to format the code a bit.

PR #374, opened 2023-04-26
Remove unnecesary max width that was creating issues on the layout of the module globally

RodrigoCastroS:


vicarali (approved):
LGTM! Nice improvement!

PR #381, opened 2023-05-09
Feature/case-studies/breadcrumbs

vicarali:
Change the static Industry tag on Case Study pages for the same breadcrumbs menu used on the Industry pages


version_control/Codurance_September2020/css/_menus.css
```diff
@@ -32,6 +32,10 @@ menu items appear, including the link and hover effects.
 }
 
 /* Breadcrumbs */
+.hs-breadcrumb-menu-wrapper {
```
RodrigoCastroS, line 35: Maybe here we can use a spacing variable instead a fix pixel number. What do you think?

version_control/Codurance_September2020/templates/case-study-template.html
```diff
@@ -18,14 +18,15 @@
...
         Case Study
       </span>
-      <p class="case-study-header__industry">
-        {% module "industry_text"
-          label="Industry Text",
-          path="@hubspot/text",
-          value="Health Sector"
-          no_wrapper=True
+
+      <div class="case-study-header__breadcrumbs">
```
RodrigoCastroS, line 22: I'm not really sure how Hubspot implements breadcrumbs but we could take in consideration accessibility. This is a [link to WAI standards](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/)

RodrigoCastroS (dismissed):
I've left you a couple of comments. We can discuss it further if you want to improve what we have. Other than that, well done! 🚀

version_control/Codurance_September2020/css/_menus.css
vicarali: I wasn't sure with doing that due to this variables being mainly to define spaces between elements, but I guess it does make sense to use it here, so I just did the change!

version_control/Codurance_September2020/templates/case-study-template.html
vicarali: Hubspot uses a div wrapper with `role="navigation"`, perhaps it is enough?

PR #384, opened 2023-05-22
Feature/nav-menu/refactoring

vicarali:
Refactoring process of the main navigation menu, including the following tasks:
- Clean code of both the menu-section module and site's header in order to keep it as simple as possible
- Update design of the submenus
- Change layout of the submenu items to have it as a grid instead of a simple list


RodrigoCastroS:
We've decided to not deploy the menu as it is. We want to include the promotional section when launching the first iteration of the menu

PR #391, opened 2023-06-09
Update/nav-menu/group-utilities

RodrigoCastroS:
The current branch holds 2 new features:

1. Now the user can choose the language from a dropdown selector and those pages wihtout translation would show the language grayed out.
2. The desktop navigation will unfold when the user hovers over any menu item with a submenu


version_control/Codurance_September2020/css/modules/menu-section.css
```diff
@@ -309,27 +321,35 @@
...
-  .navigation-primary .submenu.level-1 > li > .menu-link {
-    padding-inline: var(--space-0);
-  }
+  .has-submenu:hover .website-navigation-sub-menu {
+    visibility: visible;
+    opacity: 1;
+  } 
 
-  .website-navigation-sub-menu.website-navigation-sub-menu--open {
+  /* .website-navigation-sub-menu.website-navigation-sub-menu--open {
```
vicarali, line 329: We should be able to remove this commented block now that we use the hover display method, right? If so, please remove it from the main.js file too if there's any reference to this class on there

vicarali (approved):
LGTM! We've already fixed all the pending issues, like the accesibility adjustments, so we should be able to launch now. 

Happy to get this out of the way 🍾

PR #394, opened 2023-06-16
Create a specific selector for all languages and hide it on mobile

RodrigoCastroS:
At the moment on mobile screens the contact us link was still showing at the top bar menu on ES and PT sites. This is the fix to it applying an specific css class for all languages


vicarali (approved):
LGTM! 🚀

PR #400, opened 2023-07-10
Feature/search-bar/display-on-click

RodrigoCastroS:
On this request we implemented two new features that will improve UX after receiving feedback. Those changes would be:
* Display the search bar on click
* Hide it when clicking outside or pressing the Escape key.
* When clicking, having the cursor straight into input to start typing.

### DIFF
--- a/version_control/Codurance_September2020/css/partials/utilities-menu.css
+++ b/version_control/Codurance_September2020/css/partials/utilities-menu.css
@@ -44,6 +44,11 @@
   transition: color .5s;
 }
 
+.search-bar__button:hover {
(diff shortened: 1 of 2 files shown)


vicarali (dismissed):
I've tested the new display method on both mobile/desktop and the click/focus/keydown events work perfectly, so good job!!

I just sent you a message regarding the display animation since I find it a little bit clunky for some reason, but I'm not sure why. Other than that, LGTM!

vicarali (approved):
Once again, LGTM! Thank you for the adjustments on the transitions 🙏

PR #406, opened 2023-07-24
new-module/contact-us-page/office-listing

RodrigoCastroS:
This is part of the process of revamping the contact us page.

### DIFF
--- a/version_control/Codurance_September2020/css/modules/offices-listing.css
+++ b/version_control/Codurance_September2020/css/modules/offices-listing.css
@@ -1,0 +1,6 @@
+{% import '../utils/utils.css' as utils %}
+
+.office__text p,
+.office__text a {
+    {{ utils.natus() }}
+}
--- a/version_control/Codurance_September2020/css/modules/offices-listing.css
+++ b/version_control/Codurance_September2020/css/modules/offices-listing.css
@@ -5,12 +5,31 @@
     align-items: center;
     gap: var(--space-3);
 }
+
+.office:not(:first-child) {
+    margin-top: var(--space-2);
+}
         
-.office img {
+.office__image {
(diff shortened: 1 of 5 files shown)


version_control/Codurance_September2020/css/modules/offices-listing.css
```diff
@@ -0,0 +1,35 @@
+{% import '../utils/utils.css' as utils %}
+
+.office {
+    display: flex;
+    align-items: center;
+    gap: var(--space-3);
```
vicarali, line 6: This gap seems to be a little bit larger on the design (48px/--space-5), so maybe we should increase it here too?

version_control/Codurance_September2020/css/modules/offices-listing.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.office:not(:first-child) {
+    margin-top: var(--space-2);
```
vicarali, line 10: Same case here with this gap, on the design seems to be a little bit larger (48px/--space-5, the same value as before)

version_control/Codurance_September2020/css/modules/offices-listing.css
vicarali:
```suggestion
    gap: var(--space-5);
```

version_control/Codurance_September2020/css/modules/offices-listing.css
vicarali:
```suggestion
    margin-top: var(--space-5);
```

vicarali (approved):
I just left you a pair of comments regarding some spaces that maybe could be adjusted, but other than that good job! LGTM! 🤘

PR #407, opened 2023-07-25
Update/industries/public-sector/design-review-adjustments

vicarali:
Adjustments on Public Sector Industry template after review from design


version_control/Codurance_September2020/css/pages/industry-public-sector.css
```diff
@@ -35,23 +31,25 @@
   padding-bottom: var(--space-7);
 }
 
+.partners__header {
+  max-width: 55ch;
```
RodrigoCastroS, line 35: I wonder if we can use here the freed max-w instead a set value
```suggestion
  max-width: var(--freed-max-width);
```

version_control/Codurance_September2020/css/pages/industry-public-sector.css
```diff
@@ -65,25 +63,33 @@
...
 
-.case-studies .section-header h2 {
-  {{ utils.dudler() }}
-}
-
-.case-studies .section-header p {
-  {{ utils.freed() }}
-  font-weight: var(--light-font-weight);
+.case-studies .section-header {
+  margin-bottom: 0 !important;
```
RodrigoCastroS, line 79: In this case it seems that removing the !important rule, it would work as well.

RodrigoCastroS (approved):
It looks good to me. I've left some comments on some CSS rules but overall good 👍

version_control/Codurance_September2020/css/pages/industry-public-sector.css
vicarali: I tried, but it seems that removing it doesn't work since the default margin comes from a very specific selector on main.css 😅 

I'm not a huge fan of the important flag either but I think I will leave it here since I didn't find any selector more specific and suitable on this context

PR #409, opened 2023-08-10
Swap cta-banner and contact us section

RodrigoCastroS:
A request from Marketing UK wanting to interchange some sections of the template.


vicarali (approved):
The sections still look great after reordering, so LGTM!

PR #410, opened 2023-08-21
Update/blog-posts-newsletter/thank-you-translations

vicarali:
- Refactor all newsletter modules in order to ensure the usage of config copy
- Add copy + form configurable fields on both popup and footer newsletter forms
- Clean repeated/unused code on both popup and footer newsletter forms

### DIFF
--- a/version_control/Codurance_September2020/config/config.html
+++ b/version_control/Codurance_September2020/config/config.html
@@ -33,7 +33,7 @@
       'text': 'Nossa newsletter compartilha dicas de especialistas e estudos de caso inspiradores'
     },
     'form_id': 'f30d16f8-0126-4482-bab7-b742b70d75ed',
-    'css_class': 'cta_newsletter_spanish'
+    'css_class': 'cta_newsletter_portuguese'
   }
 }
 %}
--- a/version_control/Codurance_September2020/config/config.html
+++ b/version_control/Codurance_September2020/config/config.html
@@ -1,28 +1,22 @@
 {% set newsletter_cta = {
   'en': {
-      'pending': {
-        'header': 'Ready to be inspired?',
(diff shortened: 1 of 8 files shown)


version_control/Codurance_September2020/config/config.html
```diff
@@ -1,48 +1,36 @@
...
+      'header': 'Our newsletter shares expert tips and inspirational case studies'
+    },
+    'form_id': '91334e24-959c-4eed-80b1-e2eb60dec492'
   },
   'es': {
     'pending': {
       'header': '¿Preparado para inspirarte?',
       'text': 'Suscríbete a nuestra newsletter para que podamos hacerte llegar recomendaciones de expertos y casos prácticos inspiradores'
     },
-    'success': {
```
RodrigoCastroS, line 22: In this case it wouldn't be useful to use this object? So we can change the header and the body of the banner

RodrigoCastroS (commented):
I left you a comment that might be the main the main direction we are taking here. It seems before we used the heading and the body of the form to dynamically render the success message. Do we want to override all that?

RodrigoCastroS (commented):
I left you a comment that might be the main the main direction we are taking here. It seems before we used the heading and the body of the form to dynamically render the success message. Do we want to override all that?

RodrigoCastroS (commented):
I left you a comment that might be the main the main direction we are taking here. It seems before we used the heading and the body of the form to dynamically render the success message. Do we want to override all that?

RodrigoCastroS (approved):

RodrigoCastroS (approved):
After having a chat, it seems that the old way of displaying the message wasn't working. We decided to simplify it and reduce everything to a simple thank you message when a newsletter form is submitted.

PR #411, opened 2023-08-22
Refactor/menu-scroll-event

RodrigoCastroS:
Clean up and transform the main event listener that hides/shows the main menu (from jquery to js ES6)


vicarali (approved):
Good jQuery riddance! 😉 LGTM!

PR #426, opened 2023-10-09
Update/twitter-logo

RodrigoCastroS:
Given the fact that twitter changed their logo, we updated it on several places in our website


vicarali (approved):
LGTM! ✖️

PR #431, opened 2023-11-23
Feature/add-color-capability-to-client-logos

RodrigoCastroS:
Update a couple of features on 2 modules:

* Have the option to set color to client logos on the module.
* Change the overlay color on the industries grid overalay to a subtle linear gradient. User still can choose the color.

### DIFF
--- a/version_control/Codurance_September2020/modules/Clients-Logos.module/module.html
+++ b/version_control/Codurance_September2020/modules/Clients-Logos.module/module.html
@@ -2,18 +2,15 @@
 
 {{ require_css(get_asset_url("../../css/modules/clients-logos.css")) }}
 
-{% require_css %}
+
   <style>
-    {% scope_css %}
       .client-logos__logos-group{
         {% if module.style.logo_color == "grey" %}
-        filter: contrast(0);
+        	filter: contrast(0);
         {% endif %}    
       }
-    {% end_scope_css %}
   </style>
 
-{% end_require_css %}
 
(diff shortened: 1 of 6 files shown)


version_control/Codurance_September2020/modules/Clients-Logos.module/module.html
```diff
@@ -2,6 +2,16 @@
...
 {{ require_css(get_asset_url("../../css/modules/clients-logos.css")) }}
 
+
+  <style>
+      .client-logos__logos-group{
+        {% if module.style.logo_color == "grey" %}
+        	filter: contrast(0);
+        {% endif %}    
+      }
+  </style>
```
vicarali, line 12: I think we should wrap this style block with the scope tag in order for it to apply only on the corresponding module instance (I know we normally only use it once on each page, but just in case!) 
```suggestion
{% require_css }
  <style>
    {% scope_css }
      .client-logos__logos-group{
        {% if module.style.logo_color == "grey" %}
        	filter: contrast(0);
        {% endif %}    
      }
    {% end_scope_css }
  </style>
{% require_css }
```

vicarali (approved):

version_control/Codurance_September2020/modules/Clients-Logos.module/module.html
RodrigoCastroS: I've tried this and it doesn't work for me... Have you checked on your end?

